Issue #8, opened 2021-09-21
Use `pytest` as a testing framework


tudortimi:
`pytest` can be used for both unit testing, but also for end-to-end testing (when testing with a simulator). It leads to more readable test code.

tudortimi:
Do you have any strong preference for `nose` and `unittest`? `pytest` is by far the more popular framework.

tudortimi:
I could try to convert the existing unit tests to `pytest` to give you a glimpse.

tudortimi:
The cool thing is that `pytest` supports "classical" `unittest` tests as well, so it's not an "all or nothing" proposition. Test code can be migrated incrementally, if needed.

tudortimi:
> Does `pytest` affect running the **cocotb** testbenches using `tox` to call `make`?

Can't say for sure, but I would assume everything would still work. The unit tests (in the `nosetests` directory) are anyway separate from the cocotb tests.

raysalemi:
I've merged in your PR.  Thank you!

PR #9, opened 2021-09-21
Use pytest instead of nose

tudortimi:


tudortimi:
This will only replace the test runner.

tudortimi:
Closes #8

PR #10, opened 2021-09-22
Migrate unit tests to pytest

tudortimi:


tudortimi:
I migrated the two modules in `pytests` to fully use `pytest`. I did the changes across many small commits to that it's easier to see what's going on.

tudortimi:
I noticed that some tests in `cocotb_tests` are actually just regular `unitest` tests. They don't seem to be running in the CI at the moment. I left them alone for now, as moving them should be done in a separate PR.

raysalemi:
This is amazing!  I can see why you recommended switching.

raysalemi:
Accepting all of these changes. A *much* cleaner test structure!

PR #12, opened 2021-09-22
Add foundational register model classes

tudortimi:
Add support for foundational UVM register layer classes: `uvm_reg_block`, `uvm_reg_map`, `uvm_reg` and `uvm_reg_field`.

### DIFF
--- a/pyuvm/s18_register_model.py
+++ b/pyuvm/s18_register_model.py
@@ -26,15 +26,23 @@ class uvm_reg_map(uvm_object):
     # TODO Fix signature
     def __init__(self):
         self._parent = None
+        self._base_addr = None
 
     # 18.2.3.2
     # TODO Fix signature
-    def configure(self, parent):
+    # TODO Support binary and hex values for 'base_addr'
+    def configure(self, parent, base_addr):
         self._parent = parent
+        self._base_addr = base_addr
 
     # 18.2.4.2
     def get_parent(self):
         return self._parent
+
+    # 18.2.4.4
+    # TODO Fix signature
+    def get_base_addr(self):
+        return self._base_addr
 
 
 # 18.4.1 Class declaration
--- a/pyuvm/s18_register_model.py
+++ b/pyuvm/s18_register_model.py
@@ -3,7 +3,19 @@ from pyuvm import uvm_object
 
 # 18.1.1 Class declaration
 class uvm_reg_block(uvm_object):
-    pass
+
+    # 18.1.2.1
+    # TODO Fix signature
+    def __init__(self):
+        self._regs = []
+
+    # 18.1.3.7
+    # TODO Fix signature
+    def get_registers(self):
+        return self._regs
+
+    def _add_register(self, reg):
+        self._regs.append(reg)
 
 
 # 18.2.1 Class declaration
@@ -24,6 +36,7 @@ class uvm_reg(uvm_object):
     # 18.4.2.2
     def configure(self, parent):
         self._parent = parent
+        parent._add_register(self)
 
     # 18.4.3.1
     def get_parent(self):
--- a/pyuvm/s18_register_model.py
+++ b/pyuvm/s18_register_model.py
@@ -13,7 +13,19 @@ class uvm_reg_map(uvm_object):
 
 # 18.4.1 Class declaration
 class uvm_reg(uvm_object):
-    pass
+
+    # 18.3.2.1
+    # TODO Fix signature
+    def __init__(self):
+        self._parent = None
+
+    # 18.4.2.2
+    def configure(self, parent):
+        self._parent = parent
+
+    # 18.4.3.1
+    def get_parent(self):
+        return self._parent
 
 
 # 18.5.1 Class declaration
--- a/pyuvm/s18_register_model.py
+++ b/pyuvm/s18_register_model.py
@@ -23,12 +23,18 @@ class uvm_reg_field(uvm_object):
     # TODO Fix signature
     def __init__(self):
         self._parent = None
+        self._size = None
 
     # 18.5.3.2
     # TODO Fix signature
-    def configure(self, parent):
+    def configure(self, parent, size):
         self._parent = parent
+        self._size = size
 
     # 18.5.4.1
     def get_parent(self):
         return self._parent
+
+    # 18.5.4.3
+    def get_n_bits(self):
+        return self._size
--- a/pyuvm/s18_register_model.py
+++ b/pyuvm/s18_register_model.py
@@ -9,3 +9,8 @@ class uvm_reg_block(uvm_object):
 # 18.2.1 Class declaration
 class uvm_reg_map(uvm_object):
     pass
+
+
+# 18.4.1 Class declaration
+class uvm_reg(uvm_object):
+    pass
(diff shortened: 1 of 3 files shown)


tudortimi:
I wanted to get your feedback on a couple of things, before I continue with the work.

tudortimi:
I noticed you like to put comments with references to the standard. I added refs for both the classes and the methods. I noticed, though, that not all existing classes have the refs. What would you like to see moving forward?

tudortimi:
I noticed that you have `raise error_classes.UVMNotImplemented(...)` for methods defined in the standard, but that aren't yet implemented. There are quite a few methods in the classes I want to add. Is it ok to add the exceptions in a subsequent PR?

tudortimi:
Shall I add type annotations to method declarations?

tudortimi:
I snuck an [EditorConfig](https://editorconfig.org/) file on the branch. I'm trying to use more of it. Do you have any objections? Do you want it in a separate PR?

raysalemi:
> I noticed you like to put comments with references to the standard. I added refs for both the classes and the methods. I noticed, though, that not all existing classes have the refs. What would you like to see moving forward?

That practice went through some evolution, and I need to standardize.  I'd like to have the comment reference the spec and the docstring be useful in an IDE.  In many cases I put the standard reference in the docstring, which is useless. 

I have to admit that I don't know the first thing about the register layer. I should learn it so I can help out if you'd like a second pair of eyes.

raysalemi:
> I snuck an [EditorConfig](https://editorconfig.org/) file on the branch. I'm trying to use more of it. Do you have any objections? Do you want it in a separate PR?

Thank you!  I did not know this existed.

raysalemi:
> I snuck an [EditorConfig](https://editorconfig.org/) file on the branch. I'm trying to use more of it. Do you have any objections? Do you want it in a separate PR?

You can add it to this PR if you like. Either way works for me.

tudortimi:
> > I noticed you like to put comments with references to the standard. I added refs for both the classes and the methods. I noticed, though, that not all existing classes have the refs. What would you like to see moving forward?
> 
> That practice went through some evolution, and I need to standardize. I'd like to have the comment reference the spec and the docstring be useful in an IDE. In many cases I put the standard reference in the docstring, which is useless.
> 
> I have to admit that I don't know the first thing about the register layer. I should learn it so I can help out if you'd like a second pair of eyes.

Actually wanted to ask about docstring. So, copy it from the standard?

tudortimi:
I'd like to put the PR back into draft state, 'cause it's far from ready, but I don't know how.

raysalemi:
> Shall I add type annotations to method declarations?

I don't want to that. 

I think one of the wonders of Python is its typelessness and I make a big deal about that in my blog posts, and in my upcoming book.  Adding type annotations would confuse the teaching message.

tudortimi:
> > Shall I add type annotations to method declarations?
> 
> I don't want to that.
> 
> I think one of the wonders of Python is its typelessness and I make a big deal about that in my blog posts, and in my upcoming book. Adding type annotations would confuse the teaching message.

Type annotations are for code checking tools, not for the interpreter.

raysalemi:
BTW. I find your style of many small commits to be very helpful. I tend to squash all commits into one big commit, but I typically work alone. Your approach of showing them step-by-step is excellent!

tudortimi:
> BTW. I find your style of many small commits to be very helpful. I tend to squash all commits into one big commit, but I typically work alone. Your approach of showing them step-by-step is excellent!

We typically use "merge commit" as a merge strategy (actually rebase + merge, but we use Bitbucket). It gives the best results, IMO. Granular history, for readability + annotations, and coarse history (with 'git log --first-parent') for good overview.

raysalemi:
@tudortimi , I presume that I wait until you change the status from "Draft" before merging this PR, right?

tudortimi:
Yep, got a lot more to do. Drafts are for early feedback.

tudortimi:
Another question.

There's a function, `uvm_reg::get_fields`. Its signature is `virtual function void get_fields(ref uvm_reg_field fields[$])`. How would you handle it?

IMO, this is just a lazy substitute for the fact that SV doesn't easily allow one to return arrays. The devs should have used the tried and true idiom of defining a return type using an extra `typedef`. Such a method signature, that uses pseudo-returns, makes the client code difficult to write.

I would do:
```
def get_fields():
    result = []
    // ...
    return result
```

raysalemi:
My rule is to always refactor to use Python's power. So would do what makes sense for Python. In this case return a list.

pyuvm/s18_register_model.py
```diff
@@ -0,0 +1,85 @@
...
+        # TODO Support binary and hex values for 'reset'
+        self._parent = parent
+        self._size = size
+        self._lsb_pos = lsb_pos
+        self._access = access
+        self._is_volatile = is_volatile
+        self._reset = reset
+
+    # 18.5.4.1
+    def get_parent(self):
```
themperek, line 58: Maybe one can add:
```python3
parent = property(get_parent)
````
so it is more Python-like access avalaible?

raysalemi: I made a design decision keep the `get_*` rather than convert them to properties. 

I've never see the `property()` function, I've always used the decorator. 

What does the function do?

themperek: You can keep `get_*` I guess make sense for compatibility.

If you add `property(get_parent)` one can do `obj.parent` also and this will call `obj.get_parent()`. 

It is up to you. Bit matter of teste.

Some more: https://www.programiz.com/python-programming/property

raysalemi: Thanks for the link.  That's excellent in that it keeps the get_* and set_* that the UVM programmer expects but also supports the Python writer's instinct.  

If we add it we should do it across the board in one release so that the user isn't frustrated by finding that some set/get pairs don't have a property.

tudortimi:
This PR is already quite big, so I'd rather stop at implementing the introspection related methods. Code for values can be added subsequently.

tudortimi:
As already discussed, I would propose merging using a merge commit. (This is, of course, after any eventual changes you would like made.)

raysalemi (approved):
This all looks great!

tudortimi:
Thanks for incorporating this PR. I see you used rebase, though. This puts each commit in a "linear" path in the history. Have a look at the history for SVUnit, where I use merge commits: This way, the commits in a PR stay "together" and can be treated as a logical group. If you clone the repo and do `git log --first-parent`, you'll only see a list of PR merges, which is a good overview, but it's always possible to see the entire history with granular commits.

tudortimi:
The history on GitHub will always show all commits, unfortunately. Bitbucket is better in this respect, as it shows the lines for the merge commits as well.

raysalemi:
> Thanks for incorporating this PR. I see you used rebase, though. This puts each commit in a "linear" path in the history. Have a look at the history for SVUnit (https://github.com/svunit/svunit/commits/master), where I use merge commits: This way, the commits in a PR stay "together" and can be treated as a logical group. If you clone the repo and do `git log --first-parent`, you'll only see a list of PR merges, which is a good overview, but it's always possible to see the entire history with granular commits.

Ah. I see that now.  Thanks for the tip!  I'll do it the other way in the future.

PR #25, opened 2021-11-10
Now tests check the results.xml file

### LINKED ISSUE #19: Update tests to fail if cocotb test has FAIL in the log
Today all tests using `@cocotb.test()` pass regardless of whether they say they fail.  

**pyuvm** needs a mechanism to see the `FAIL`  messages and fail.

raysalemi:
resolves #19 

Now we use combine_results.py from cocotb to check that all tests actually pass.


raysalemi:
This should also explicitly merge all commits that got squashed out earlier.

PR #27, opened 2021-11-11
Fix DeprecationWarning

cjdrake:
Was seeing this:

pyuvm/examples/TinyALU/tinyalu_cocotb.py:69:
DeprecationWarning: Setting values on handles
using the ``dut.handle = value`` syntax is deprecated.
Instead use the ``handle.value = value`` syntax

Also get rid of discouraged glob import style.
According to PEP8, "Wildcard imports ... should be avoided".

### DIFF
--- a/examples/TinyALU/.gitignore
+++ b/examples/TinyALU/.gitignore
@@ -2,3 +2,5 @@ sim_build
 __pycache__
 results.xml
 
+# VCS outputs
+ucli.key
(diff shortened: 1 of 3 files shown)


raysalemi:
Hi Chris,

I've been meaning to remove the deprecated use of `<=`, so thank you.  Also, glad to have VCS files in the .gitignore.

As for the `*` import.  I can see removing it from `tinyalu_uvm` but starred import is standard UVM usage, so I want to keep it in `from pyuvm import *`. 

Is there a way to accept some hunks and not others, or do you need to modify the pull request?  I'm new to GitHub project management.

Ray

cjdrake:
Thanks for looking at it. I'll fix the commit, and re-upload.

cjdrake:
Reverted changes related to glob imports, in code and commit log.

ptal.

raysalemi (approved):
Thanks for the fixes, Chris!  It's interesting that flake8 is happy whether or not return has parentheses.

PR #36, opened 2021-12-07
Default to parent's logging level, if any

jonpovey:
This allows e.g. enabling debug level messages during build_phase of
child components.

Use case, approx:
```
class MyTest(uvm_test):
    def build_phase(self):
        self.set_logging_level_hier(logging.DEBUG)
        self.logger.debug('in build phase')
        self.env = MyEnv.create('env', self)

class MyEnv(uvm_env):
    def build_phase(self):
        self.logger.debug('I want this to be seen')
```

### DIFF
--- a/pyuvm/s06_reporting_classes.py
+++ b/pyuvm/s06_reporting_classes.py
@@ -29,13 +29,17 @@ class PyuvmFormatter(SimColourLogFormatter):
 class uvm_report_object(uvm_object):
     """ The basis of all classes that can report """
 
-    def __init__(self, name):
+    def __init__(self, name, parent=None):
         super().__init__(name)
         uvm_root_logger = logging.getLogger('uvm')
         # Every object gets its own logger
         logger_name = self.get_full_name() + str(id(self))
         self.logger = uvm_root_logger.getChild(logger_name)
-        self.logger.setLevel(level=logging.INFO)  # Default is to print INFO
+        # Default to parent's logging level, else INFO
+        if parent is not None:
+            self.logger.setLevel(level=parent.logger.level)
+        else:
+            self.logger.setLevel(level=logging.INFO)
         # We are not sending log messages up the hierarchy
         self.logger.propagate = False
         self._streaming_handler = logging.StreamHandler(sys.stdout)
--- a/pyuvm/s13_uvm_component.py
+++ b/pyuvm/s13_uvm_component.py
@@ -44,7 +44,7 @@ is really going on.  We've opted for the latter.
         if parent is not None:
             parent.add_child(name, self)
         self.print_enabled = True  # 13.1.2.2
-        super().__init__(name)
+        super().__init__(name, parent)
 
         # Cache the hierarchy for easy access
         if name != 'uvm_root':


raysalemi:
Hi Jon,
I see the problem you're solving here, and you're right that the ability to set a global logging level is missing in **pyumv**.  SystemVerilog UVM solves this problem with a command-line option (+UVM_VERBOSITY), but we don't use the command line in **pyuvm**.

Rather than violate 1800.2 by modifying `uvm_report_object` I'd rather that we use the `ConfigDB()` to solve this. I recommend that we define a `ConfigDB()` label as `UVM_LOGGING_LEVEL` and then your code would look like this:

```
class MyTest(uvm_test):
    def build_phase(self):
        self.logger.debug('in build phase')
        self.env = MyEnv.create('env', self)

class MyEnv(uvm_env):
    def build_phase(self):
        ConfigDB().set(None, "*", "UVM_LOGGING_LEVEL", DEBUG)
        self.logger.debug('I want this to be seen')
```

Then the modification to `uvm_report_object` would be like this:

```
    def __init__(self, name, parent=None):
         super().__init__(name)
         uvm_root_logger = logging.getLogger('uvm')
         # Every object gets its own logger
         logger_name = self.get_full_name() + str(id(self))
         self.logger = uvm_root_logger.getChild(logger_name)
         self.logger.setLevel(level=logging.INFO)  # Default is to print INFO
         # Default to parent's logging level, else INFO
         try:
             logging_level = ConfigDB().get(self, "", "UVM_LOGGING_LEVEL")
             self.logger.setLevel(level=logging_level)
         except UVMConfigItemNotFound:   # (sp?)
             self.logger.setLevel(level=logging.INFO)
```

raysalemi:
@jonpovey , what do you think of this solution?

```
import cocotb
from pyuvm import *


class MyComp(uvm_component):
    def build_phase(self):
        self.logger.debug("In Comp")

    async def run_phase(self):
        self.raise_objection()
        self.drop_objection()


class ChainTest(uvm_test):

    def build_phase(self):
        self.logger.debug("My debug message in test")
        self.comp = MyComp("comp", self)


@cocotb.test()
async def chain_test(_):
    """Demonstrate default logging levels"""
    uvm_report_object.set_default_logging_level(DEBUG)
    await uvm_root().run_test("ChainTest")
```

output

```
     0.00ns INFO     Starting test: "chain_test"
                     Description: Demonstrate default logging levels
     0.00ns DEBUG    testbench.py(17)[uvm_test_top]: My debug message in test
     0.00ns DEBUG    testbench.py(7)[uvm_test_top.comp]: In Comp
```

raysalemi:
I've implemented the default logging function above and tested it. So I'm closing this PR.

jonpovey:
That should work fine for my purposes, thanks. I may continue to throw half-assed patches at you, if you are happy to police the spec-compliance.
Also FYI you are unlikely to get fast responses from me during your working day as I am in Brisbane, Australia on UTC+10

raysalemi:
Yes. Please continue to contribute. I need the viewpoints of others.

PR #43, opened 2021-12-08
Move default logging level config out of uvm_report_object

jonpovey:
Move logging config into a module dict, similar to matplotlib rcParams.
This avoids cluttering the namespace of objects derived from
uvm_report_object.

Example usage:
```
async def generic_runner(dut):
    pyuvm_logging_config['default_level'] = logging.DEBUG
```

### DIFF
--- a/pyuvm/s06_reporting_classes.py
+++ b/pyuvm/s06_reporting_classes.py
@@ -12,6 +12,9 @@ import sys
 from cocotb.log import SimColourLogFormatter, SimTimeContextFilter
 from logging import DEBUG, CRITICAL, ERROR, WARNING, INFO, NOTSET, NullHandler   # noqa: F401, E501
 
+pyuvm_logging_config = {
+    'default_level': logging.INFO,
+}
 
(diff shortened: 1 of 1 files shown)


jonpovey:
I noticed (while trying to add more options to logging config for my own amusement) that adding more attributes to the class meant that all derived classes and objects were going to inherit these attributes, and there might be clashes later.
It seemed like they should be more "off to the side", this patch takes them out of the class namespace.

raysalemi:
Hi @jonpovey ,

I'm going to pass on this PR for a couple of reasons:

1. The `ConfigDB()` singleton almost always handles sharing information across the testbench, and one can usually set the `ConfigDB()` before calling `uvm_root().run_test()`.
2. This use model doesn't follow typical UVM usage, which relies upon `set_` `get_` accessors because of its SystemVerilog heritage.

That said, there is some entanglement between the `uvm_report_object` and `ConfigDB()`.  The `ConfigDB()` calls `uvm_report_object` and creates a logger so that it can log debug messages.  It might bear refactoring. 

Thanks for your help and advice!

Ray

PR #48, opened 2021-12-09
phasing: replace fork() with start_soon()

jonpovey:
fork() is deprecated, also it causes a simulator crash if a forked
coroutine raises an exception before await-ing.
See https://github.com/cocotb/cocotb/issues/2804


jonpovey:
This commit does not replace any of the various `fork()` calls in tests / examples.

cmarqu:
This probably requires cocotb 1.6 then in https://github.com/pyuvm/pyuvm/blob/master/setup.py#L27

raysalemi:
@jonpovey, please add the suggested changes to setup.py so that we require cocotb 1.6.

raysalemi:
I changed the requirement.  Thanks for letting me know about this.

PR #57, opened 2021-12-13
Objection fixes

jonpovey:
Test cases and fixes for a couple of objection issues.
The main one being that a crashing test would leave raised objections, causing a following test to never end.

I added a "test" to see if a warning is printed, but it doesn't actually test if the warning gets printed - I don't know how to do that. It needs some kind of mocking of the logger, I guess.


raysalemi:
Great catch!  I hadn't thought of multiple runs in a file when I wrote this initially.

PR #69, opened 2021-12-20
More helpful error when port is not connected

jonpovey:
A minor modification for your consideration:

Print a more helpful error using an assertion if a driver tries to
get_next_item before it has been connected to an export.
Do this in a try/except block to avoid the cost of checking every time.

Changes this:
```
AttributeError: 'NoneType' object has no attribute 'get_next_item'
```
into this:
```
AssertionError: export is not connected
```

### DIFF
--- a/pyuvm/s14_15_python_sequences.py
+++ b/pyuvm/s14_15_python_sequences.py
@@ -237,7 +237,11 @@ class uvm_seq_item_port(uvm_port_base):
     async def get_next_item(self):
         """get the next sequence item from the request queue
         """
-        return await self.export.get_next_item()
+        try:
+            return await self.export.get_next_item()
+        except AttributeError:
+            assert self.export is not None, "export is not connected"
+            raise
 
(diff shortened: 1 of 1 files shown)


raysalemi (changes requested):
It's a good idea. Could you remove the `e`? I don't believe it's needed.

```python
#  example code 
try:
    x = 3/0
except ZeroDivisionError:
    print("Oops")
    raise

# result

Oops
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-1-4a24820244f5> in <module>
      1 try:
----> 2     x = 3/0
      3 except ZeroDivisionError:
      4     print("Oops")
      5     raise

ZeroDivisionError: division by zero

```

jonpovey:
> It's a good idea. Could you remove the `e`? I don't believe it's needed.

You're right. Done.

raysalemi (approved):

PR #71, opened 2021-12-21
give uvm_sequence a default name

jonpovey:
As discussed


jonpovey:
Pushed a little bonus cleanup into the same branch because I'm lazy.

PR #90, opened 2022-02-12
Add pyuvm.test decorator

### LINKED ISSUE #89: Simplify making tests
From the example I see the following code repeated often.

```python
@cocotb.test()
async def alu_test(_):
    """Test ALU with random and max values"""
    await uvm_root().run_test(AluTest)
```

It seems like you need to manually write this stub for every test. This could perhaps be simplified by adding a metaclass to `uvm_test` to instrument tests so they behave as coroutines and can be directly decorated with `@cocotb.test()` or use in `TestFactory`. In the end it would look like the following.

```python
@cocotb.test(timeout=1, timeout_unit='ns')
class AluTest(uvm_test):
    ...
```

ktbarrett:
This simplifies the process of creating tests. Instead of having to
create a coroutine stub that calls uvm_root().run_test(), this is done
for the user with a decorator that directly wraps the uvm_test class
definition.

Closes #89.

### DIFF
--- a/examples/TinyALU/testbench.py
+++ b/examples/TinyALU/testbench.py
@@ -255,7 +255,10 @@ class AluEnv(uvm_env):
         self.driver.ap.connect(self.scoreboard.result_export)
 
 
+@test()
 class AluTest(uvm_test):
+    """Test ALU with random and max values"""
+
     def build_phase(self):
         self.env = AluEnv("env", self)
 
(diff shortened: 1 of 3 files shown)


raysalemi:
Wow. This looks great, @ktbarrett !

The only problem is that it is failing a tox test that expects an error.  The previous version passes because I tell cocotb.test to expect the error.

Is that instruction not getting passed along, or is it just not there?

raysalemi:
Strange. Now it passes.  A glitch I guess

raysalemi:
@ktbarrett just saw your message.  *That's* why its passing.

All set!

PR #92, opened 2022-02-13
Fix issue with stage on test decorator

ktbarrett:
The stage variable on the test decorator defaults to None, which is broken and we are changing it upstream. But until that percolates the default here will have to stay `None` to prevent issues when `cocotb.test` and `pyuvm.test` are used in the same module and the stages on each are `None` and `0`, which fail to compare.


raysalemi (approved):

raysalemi:
Thanks!

PR #108, opened 2022-08-04
Add TLM methods to uvm_tlm_fifo_base

### LINKED ISSUE #107: Add TLM methods to uvm_tlm_fifo_base
In SystemVerilog UVM, FIFOs can call TLM methods. For example,
```systemverilog
uvm_tlm_analysis_fifo #(my_seq_item) my_fifo;
...
my_fifo.get(item);
```
In pyuvm, this requires an extra step of attaching a get port and using its get method.
```python
self.my_fifo = uvm_tlm_analysis_fifo("my_fifo", self)
self.my_get_port = uvm_get_port("my_get_port", self)
...
self.my_get_port.connect(self.my_fifo.get_export)
...
item = await self.my_get_port.get()
```

However, `uvm_tlm_fifo_base` could implement the TLM methods as follows:
```python
async def get(self):
    return await self.get_export.get()
```

Then FIFOs would work like in SystemVerilog.
```python
item = await self.my_fifo.get()
```

miserva:
Adds possibility to call TLM methods directly with FIFOs.
```python
self.my_fifo = uvm_tlm_analysis_fifo("my_fifo", self)
item = await self.my_fifo.get()
```
Tests probably need updating but I cannot run them in my environment.

Closes #107

### DIFF
--- a/pyuvm/s12_uvm_tlm_interfaces.py
+++ b/pyuvm/s12_uvm_tlm_interfaces.py
@@ -685,6 +685,33 @@ class uvm_tlm_fifo_base(uvm_component):
                                                                              self.queue, self.get_ap)  # noqa: E501
         self.get_peek_export = self.uvm_GetPeekExport("get_peek_export", self, self.queue, self.get_ap)  # noqa: E501
 
+    async def put(self, item):
+        await self.put_export.put(item)
+
+    def can_put(self):
+        return self.put_export.can_put()
+
+    def try_put(self, item):
+        return self.put_export.try_put(item)
+
+    async def get(self):
+        return await self.get_export.get()
+
+    def can_get(self):
+        return self.get_export.can_get()
+
+    def try_get(self):
+        return self.get_export.try_get()
+
+    async def peek(self):
+        return await self.peek_export.peek()
+
+    def can_peek(self):
+        return self.peek_export.can_peek()
+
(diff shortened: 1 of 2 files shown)


raysalemi:
> Adds possibility to call TLM methods directly with FIFOs.
> 
> ```python
> self.my_fifo = uvm_tlm_analysis_fifo("my_fifo", self)
> item = await self.my_fifo.get()
> ```
> 
> Tests probably need updating but I cannot run them in my environment.
> 
> Closes #107

Is there a problem with the tests or testing system? 

This pull request passes the tests but fails the flake8 linting check.  Can you run flake8?  (pip install flake8)

Final point.  Shouldn't we implement calls for all the exports in TLM FIFO including transport?

miserva:
The problem is with my system setup specifically. I cannot run EDA tools in a virtual environment.

It seems the flake8 issue was unrelated to this PR and had to do with a new flake8 release. I added a fix anyway.

I'm not sure I understand your last comment. FIFOs don't seem to have transport. `fifo.transport(..)` does not work in SV UVM either. Do you mean adding `transport` and `nb_transport` calls to `uvm_tlm_transport_channel`? I'm not familiar with that class at all.

raysalemi:
To be honest, I never checked the fifo for transport.  I just made an assumption.  Bad on me!

PR #117, opened 2022-10-12
Fix test decorator to work with cocotb 1.7

ktbarrett:
Should fix the issue mentioned [here](https://gitter.im/cocotb/Lobby?at=6343f355773e7e5c7bcaa167).

### DIFF
--- a/pyuvm/extension_classes.py
+++ b/pyuvm/extension_classes.py
@@ -15,7 +15,6 @@ def test(
     stage=None,
 ):
     version_info = tuple(int(n) for n in cocotb.__version__.split("."))
-    print(version_info)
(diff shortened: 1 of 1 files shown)


pyuvm/extension_classes.py
```diff
@@ -14,6 +14,11 @@ def test(
     skip=False,
     stage=None,
 ):
+    version_info = tuple(int(n) for n in cocotb.__version__.split("."))
+    print(version_info)
```
ktbarrett, line 18:
```suggestion
```

ktbarrett:
>Is the print statement on line 18 a debug statement?

Whoops...

PR #118, opened 2022-10-13
Fix #112, run_phase now waits for objections from all components

### LINKED ISSUE #112: Run phase doesn't wait for objections from all components to drop
If there are two (or more) components which raise and drop objections in a test, the first component which reaches zero remaining raised objections will end the run phase, even if other components have not yet dropped all of their objections.

For example, suppose we have two components c1 and c2:
```
c1 raise_objection
c2 raise_objection
c1 drop_objection
<<< run phase erroneously ends here
c2 drop_objection
<<< run phase should end here
```

I believe the bug is in this code in utility_classes.py:
```python
    def raise_objection(self, raiser):
        name = raiser.get_full_name()
        self.__objections[name] = self.__objections.setdefault(name, 0) + 1
        self.objection_raised = True
        self._objection_event.clear()

    def drop_objection(self, dropper):
        name = dropper.get_full_name()
        try:
            self.__objections[name] -= 1
        except KeyError:
            self.objection_raised = True
            pass
        if self.__objections[name] == 0:
            del self.__objections[name]
            self._objection_event.set()
```

Note that the `_objection_event` is set if only `name`'s objections are all dropped, without checking if any other objections in `__objections` are raised.

I think a simple modification like this might suffice:
```python
    def drop_objection(self, dropper):
        name = dropper.get_full_name()
        try:
            self.__objections[name] -= 1
        except KeyError:
            self.objection_raised = True
            pass
        if self.__objections[name] == 0:
            del self.__objections[name]
            
            # only signal all objections done if none exist anywhere
            if len(self.__objections) == 0:
                self._objection_event.set()
```

I think existing tests in pyuvm could be modified in minor ways to cover this.

I discovered the issue in reimplementing some code originally written in ROHD-VF (https://github.com/intel/rohd-vf) to compare functionality and features with pyuvm.

mkorbel1:
Fix #112, bug where `run_phase` could end before all objections were dropped from all components.  Also added a new test to protect functionality.  Confirmed test fails without the fix and passes with it.

### DIFF
--- a/pyuvm/utility_classes.py
+++ b/pyuvm/utility_classes.py
@@ -241,7 +241,10 @@ class ObjectionHandler(metaclass=Singleton):
             pass
         if self.__objections[name] == 0:
             del self.__objections[name]
-            self._objection_event.set()
+
+            # only signal all objections done if none exist anywhere
+            if len(self.__objections) == 0:
+                self._objection_event.set()
 
     async def run_phase_complete(self):
         # Allow the run_phase coros to get scheduled and raise objections:
--- a/tests/cocotb_tests/run_phase/test.py
+++ b/tests/cocotb_tests/run_phase/test.py
@@ -15,9 +15,11 @@ class my_error(uvm_test):
         self.raise_objection()
         raise UVMError("This is an error")
 
+
 class my_no_objection(uvm_test):
     async def run_phase(self):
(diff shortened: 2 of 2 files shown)


raysalemi (approved):

raysalemi:
Thank you so much!

PR #132, opened 2023-01-20
Fix colored output when COCOTB_ANSI_OUTPUT is set 0

viewtifulRafa:
pyuvm colored output can not be disabled as it always uses the _SimColourLogFormatter_

For post-processing of log files normally is desirable to have a plain log file without colors. For this purpose cocotb provides the COCOTB_ANSI_OUTPUT Environment variable.

Fix adds support for the COCOTB_ANSI_OUTPUT Environment variable to pyuvm logging.

### DIFF
--- a/pyuvm/s06_reporting_classes.py
+++ b/pyuvm/s06_reporting_classes.py
@@ -9,7 +9,8 @@
 from pyuvm.s05_base_classes import uvm_object
 import logging
 import sys
-from cocotb.log import SimColourLogFormatter, SimTimeContextFilter
+from cocotb.log import SimLogFormatter, SimColourLogFormatter, SimTimeContextFilter
+from cocotb.utils import want_color_output
 from logging import DEBUG, CRITICAL, ERROR, WARNING, INFO, NOTSET, NullHandler   # noqa: F401, E501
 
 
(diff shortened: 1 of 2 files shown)


raysalemi:
Thank you!

I've enabled regressions for you.  They failed this time.

viewtifulRafa:
Thanks!
I will review the failures.

viewtifulRafa:
I fixed the errors and tox and flake are passing on my side.
I have pushed the changes to my fork.

raysalemi:
Great!

What was the tox error?

Thanks,
Ray

amal-khailtash:
This is a good solution.  In general, if a script outputs ANSI, it should have an option or when redirected to a file, to not output ANSI color codes at all.  The logging to file should have this option as well.

Normally if a script/tool does not support this, I use this handy alias to remove ANSI codes from a log file or as a pipe on terminal:

`alias strip_ansi='sed -r "s/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[m|K]//g"'`

raysalemi:
Thank you for this enhancement!

viewtifulRafa:
Hi Ray,
tox was failing because it does not allow external commands by default and was reporting and error when trying to call make.
I added make to the tox config file in the allowlist_external field and it resolved the issue.

raysalemi:
Interesting. That seems like new behavior.

PR #133, opened 2023-02-05
Change "UVMNotImplemented" error to "UsePythonMethod" error  when applicable

hzisman:


raysalemi:
Did I create the `UsePythonMethod` exception and then not use it?

raysalemi:
The test failed. I wonder if it expects the previous exception.

hzisman:
> Did I create the UsePythonMethod exception and then not use it?

You use, but there are still cases you use "UVMNotImplemented" when It should be "UsePythonMethod"

hzisman:
> The test failed. I wonder if it expects the previous exception.

Yes, it does 
for example: https://github.com/pyuvm/pyuvm/blob/eecfb4be7ee08350d02c8a1ca90b9e74f4e4c32f/tests/pytests/test_05_base_classes.py#L107-L108
I'll send changes for the tests as well

raysalemi:
Thank you!

PR #146, opened 2023-07-11
Modify uvm_agent to use @property for active status

tj-scherer:
This looks like a perfect spot to use the @property decorator and follow [PEP8 guidelines for inheritance](https://peps.python.org/pep-0008/#designing-for-inheritance)


raysalemi:
Hi TJ,
Thank you for contributing to **pyuvm**.  

I gave this question a lot of thought when I implemented the IEEE 1800.2 standard. I agree with you in principle that the `@property` is more Pythonic and makes better code.  But the code would not look like UVM code, so I chose to implement the 1800.2 standard.

There may be room to create properties that build upon the standard.  I forget, but I think that `is_active` is specified.

As it is, this PR is failing regressions.  I'm afraid I don't have time to understand why at this moment.

Ray

tj-scherer:
I see! That makes perfect sense, thank you for clarifying.

PR #148, opened 2023-07-14
Automatic string conversion for log messages

tj-scherer:
Exception occurs when passing non-strings to the logger, this should let it print whatever comes in (as long as it has a string representation).


raysalemi:
That is a great idea.  And it's funny to see that I used the same f-string approach on line 27.  So this is more consistent as well.

PR #158, opened 2023-08-06
first PYUVM Register Model release and Regression Fix

EngRaff92:
## First PYUVM Register Model release 
pyuvm/s18_pyuvm_reg_block.py
delete mode 100644 pyuvm/s18_register_model.py
create mode 100644 pyuvm/s19_pyvum_reg_field.py
create mode 100644 pyuvm/s20_pyuvm_reg.py
create mode 100644 pyuvm/s21_pyuvm_reg_map.py
create mode 100644 pyuvm/s22_pyuvm_mem.py
create mode 100644 pyuvm/s23_pyuvm_reg_item.py
create mode 100644 pyuvm/s24_pyuvm_reg_includes.py
create mode 100644 pyuvm/s25_pyuvm_adapter.py
create mode 100644 pyuvm/s26_pyuvm_predictor.py
create mode 100644 pyuvm/s27_pyuvm_reg_pkg.py
delete mode 100644 tests/pytests/test_18_register_model.py
create mode 100755 tests/pytests/test_pyuvm_ral.py
create mode 100755 tests/pytests/test_pyuvm_reg.py
create mode 100755 tests/pytests/test_pyuvm_reg_block.py
create mode 100755 tests/pytests/test_pyuvm_reg_field.py
create mode 100755 tests/pytests/test_pyuvm_reg_map.py

PR #162, opened 2023-08-08
Use 1800.2 naming convention for all base classes

crteja:
This PR updates the naming convention as discussed in https://github.com/pyuvm/pyuvm/pull/158


raysalemi:
naming update merged into `ral_dev`

PR #173, opened 2023-08-19
Regenerate reg block using peakrdl-verilog

crteja:
- Regenerated regblock from rdl file using [peakrdl-verilog](https://github.com/hughjackson/PeakRDL-verilog)
- This implementation doesn't use packed structures that were unsupported in `icarus` and `verilator`
- Also, the reg block interface is now a simple handshake interface similar to APB but without `sel` and `enable` signals
- for sake of simplicity, register field write enable bits are always driven with 1
- Added an agent and reg adapter to simplify testing of TinyALU_reg

Assuming this module now compiles with all EDA tools since it compiles and runs with `icarus` and `verilator`


raysalemi:
This PR failed due to conflicts, perhaps with #171

raysalemi:
There are conflicts in this PR.

raysalemi:
Thanks!

PR #175, opened 2023-09-08
Fix pyuvm#160, replace all cocotb.fork() with cocotb.start_soon() in all tests

MehanathanS:
Fix Issue pyuvm#160 by Replacing all the instances of cocotb.fork() to cocotb.start_soon() in all tests


raysalemi:
Thank you!

PR #178, opened 2023-09-12
Add default argument to ConfigDB().get()

timothyscherer:
Implements changes discussed in #177.

### DIFF
--- a/pyuvm/s13_uvm_component.py
+++ b/pyuvm/s13_uvm_component.py
@@ -535,17 +535,18 @@ class ConfigDB(metaclass=utility_classes.Singleton):
 
         self.trace("SET", context, inst_name, field_name, value)
 
-    def get(self, context, inst_name, field_name, default = default_get):       
+    def get(self, context, inst_name, field_name, default=default_get):
         """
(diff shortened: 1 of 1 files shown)


raysalemi:
Why is default_get set to object() rather than None at the start?

timothyscherer:
Because someone may want to intentionally use `None` as their default return value, which is a global constant to the same exact object. In order to catch the object, the default needs to be a completely unique object (`is` uses `id(obj)` which is unique to that object aka a pointer) that the user will never pass as a default value.

By using `object()`, it is guaranteed that the exception will only get raised if no value for `default` was given.

raysalemi:
That makes sense. 

Does using the `is` remove the need to use the id?

timothyscherer:
The "under-the-hood" implementation of `is` compares `id()` of both objects, so it checks to see if they are the same object. These changes only use `is`.

`if default is self.default_get` is  functionally equivalent to `if id(default) == id(self.default_get)`.

raysalemi:
Yes. The `is` is just cleaner and more Pythonic.

raysalemi:
Thanks for the improvement!

PR #196, opened 2024-03-01
Use int instead of string for storing RAL values

KasperHesse:
This removes the functions _int2str and _hex2int that were previously used to store register addresses. 
Instead, values are now stored as plain integers, and a call to hex() is used when printing register maps with __str__

### DIFF
--- a/pyuvm/s20_uvm_reg.py
+++ b/pyuvm/s20_uvm_reg.py
@@ -22,7 +22,7 @@ class uvm_reg(uvm_object):
         self._sum: int = 0
         self._name: str = name
         self._header: str = name + " -- "
-        self._address: str = "0x0"
+        self._address: int = 0
         self._path: str = ""
         self._width = reg_width
         # If set those 2 flags will override fields values, and if set to True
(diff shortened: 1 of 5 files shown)


KasperHesse:
As noted in https://github.com/pyuvm/pyuvm/discussions/161#discussioncomment-7694613, the choice of storing addresses and offsets with strings instead of integers is counterproductive. This PR reverts that, using plain integers instead.

raysalemi:
@EngRaff92 , what do you think of this pull request?

raysalemi:
@EngRaff92   Thoughts on this?

EngRaff92:
@raysalemi apologies for this late reply I have been quite busy at work. I have to rerun the full set of tests anyway which might be impacted by the change. I do agree in having the int instead of string (the initial choice was for readability and thus the reason of the utility support function coming along) ideally I should be using the naming tuple for each register so that future generator tool could better approach this. Anyway no hassle in having them as int. I will rerun sorry again for being late

raysalemi:
I don't see that this PR has passed the tests.

PR #199, opened 2024-04-15
Add Pythonic method for raising and dropping objections

timothyscherer:
Adding this method allows users to raise objections using `with`, meaning that on any break from the loop, the objection will always be dropped.

Example usage:
```py
async def run_phase(self):
    with self.objection():
        print("run phase started")
        assert False
```

This also saves users from needing to debug a missing drop_objection, similar to Python's file opening and closing system.


raysalemi:
Thanks!

PR #212, opened 2024-06-27
Fix issues with current 2.0 support

ktbarrett:
* `cocotb.log` module was renamed to `cocotb.logging` module to prevent shadowing the `cocotb.log` variable.
* `cocotb.test` now works in a different way so we don't have to put the Test object into the module. This should also make `TESTCASE` work correctly.

### DIFF
--- a/pyuvm/_utils.py
+++ b/pyuvm/_utils.py
@@ -1,0 +1,9 @@
+import cocotb
+
+_cocotb_version_info = []
+for xx in cocotb.__version__.split("."):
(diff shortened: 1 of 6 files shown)


ktbarrett:
Guess I need version ifdefs. whoops. I'll work on this later tonight.

raysalemi:
I have modified the `pyuvm.test()` decorator.  It now accepts arguments to keep all or some singletons between tests.

Issue #214, opened 2024-06-28
uvm_report_object init gets slow for generated tests


LeoGoTi:
Hi, I have a performance-related problem on `uvm_report_object`.

It's about the initialization, the logger name, and the `setLevel` call.

## Steps to reproduce
It's simple. I use a while loop to `@pyuvm.test()`'d a test class multiple times.
The test have relatively stable behavior. I mean in the profiling `.svg` outputs, they share the similar number of calls on methods.

The problem is, the simulation speed gets slower per test. See this screenshot.
<img width="665" alt="image" src="https://github.com/pyuvm/pyuvm/assets/34080900/6e63b72f-bab8-4146-aafc-4731552aeaa9">

When test id is `02`, the `setLevel()` costs:
<img width="152" alt="image" src="https://github.com/pyuvm/pyuvm/assets/34080900/29854629-22ec-4c89-8a3b-4de4c95a8db0">

While test id grows to `39`, the `setLevel()` costs:
<img width="186" alt="image" src="https://github.com/pyuvm/pyuvm/assets/34080900/01978a51-8e9c-4047-be49-7b8da143b9bd">

## Locate the hotspot
I found that the slow-down comes from the `setLevel` call on the object's logger.
```python
class uvm_report_object(uvm_object):
    __default_logging_level = logging.INFO
    """ The basis of all classes that can report """
    def __init__(self, name):
        super().__init__(name)
        uvm_root_logger = logging.getLogger('uvm')
        # Every object gets its own logger
        logger_name = self.get_full_name() + str(id(self))
        self.logger = uvm_root_logger.getChild(logger_name)
        self.logger.setLevel(
            level=uvm_report_object.get_default_logging_level())
```

I have tried to remove the `setLevel` call, and another way to make the `logger_name` to `self.get_full_name()`, without the id-string suffix.

Only the latter helps to restore to my expected speed, but causing the output format strange (with multiple duplicates of  the `get_full_name()` string wrapped in brackets).

## My question
I am curious about the reason for making the logger name one with id-string suffix.
- What is the actual need for per-instance logger instead of per-`get_full_name()` logger?
- What is the root cause of the strange duplicate string when I have removed the id-string suffix?
- Is there a way to workaround or to fix?

raysalemi:
I don't understand what you're saying here:

>Only the latter helps to restore to my expected speed, but causing the output format strange (with multiple duplicates of the get_full_name() string wrapped in brackets).

[...]

PR #221, opened 2024-07-01
Added get_initial_logger_name to uvm_report_object

raysalemi:
I added `self.get_initial_logger_name()` to the `uvm_report_object` class.

The user can override this method to make their initial logger name.


raysalemi:
Closes issue #214

PR #225, opened 2024-08-27
fix: predicton in uvm_reg_field

hurisson:
Hello. I ran into a problem when using the automatic predictor in uvm_reg.  So, for example, I have a SCR__CTRL register containing two fields ARM [0] [RW] and RESERVED [31:1][RO]. When using uvm_reg.predict (0xFFFFFFFF, access_e.UVM_WRITE), I expect to get the value 0x1 from reg.get_mirrored_value(), but I get 0xFFFFFFFF. The problem is that an value is passed for each field in the register and for RW/WO it is not trimmed to the size of the field.  When predicting the ARM field, in its field.get_value we get 0xFFFF_FFFF, but we should get 0x1, since the field is one-bit. This received value overwrites the upper reserved bits [31:1] when calling reg.get_mirrored_value()

Here is an example of a verification scenario:

```
from pyuvm import *
import logging
import cocotb
class SCR__CTRL (uvm_reg):
    def __init__(self, name = "SCR__CTRL", reg_width = 32):
        super().__init__(name, reg_width)

    def build(self):
        self.ARM = uvm_reg_field("ARM")
        self.RESERVED = uvm_reg_field("RESERVED")
        self.ARM.configure(self, 1, 0, "RW", 1, 0x0)    
        self.RESERVED.configure(self, 31, 1, "RO", 1, 0x0)
        self.has_reset = 1

class REGBLOCK (uvm_reg_block):
    def __init__(self, name = "REGBLOCK"):
        super().__init__(name)
        self.default_map = uvm_reg_map('reg_map')
        self.default_map.configure(self, "0x0")
        self.CTRL = SCR__CTRL("CTRL")
        self.CTRL.configure(self, "0x10", "", False, False)
        self.default_map.add_reg(self.CTRL, "0x0", "RW")

reg_block = REGBLOCK("reg_block")
map       = reg_block.default_map
reg_dict  = map.get_registers()
map.reset("HARD")

for reg in reg_dict:
    if reg.get_access_policy() == "RW":
        write_data = 0xFFFF_FFFF
        reg.set_prediction(predict_t.PREDICT_DIRECT)
        reg.predict(write_data, access_e.UVM_WRITE)
        data_from_predict  = reg.get_mirrored_value()
        print(f"Write  data for register {reg.get_name()}  - {hex(write_data)} and prediction is  {hex(data_from_predict)}")
        for field in reg.get_fields():
            print(field)   
        assert data_from_predict == 0x1, f'data_from_predict {hex(data_from_predict)} is unvalid, expected 0x1'
```

image

image
image

### DIFF
--- a/pyuvm/s19_uvm_reg_field.py
+++ b/pyuvm/s19_uvm_reg_field.py
@@ -358,7 +358,7 @@ class uvm_reg_field(uvm_object):
         if self.get_access() == "RO":
             self._field_mirrored = self._reset
         if self.get_access() == "RW":
-            self._field_mirrored = wr_val & _mask
+            self._field_mirrored = wr_val
         if self.get_access() == "RC":
             self._field_mirrored = self._field_mirrored
         if self.get_access() == "RS":
@@ -396,7 +396,7 @@ class uvm_reg_field(uvm_object):
         if self.get_access() == "W0CRS":
             self._field_mirrored = self._field_mirrored & wr_val
         if self.get_access() == "WO":
-            self._field_mirrored = wr_val & _mask
+            self._field_mirrored = wr_val
         if self.get_access() == "WOC":
             self._field_mirrored = 0
         if self.get_access() == "WOS":
--- a/pyuvm/s19_uvm_reg_field.py
+++ b/pyuvm/s19_uvm_reg_field.py
@@ -358,7 +358,7 @@ class uvm_reg_field(uvm_object):
         if self.get_access() == "RO":
             self._field_mirrored = self._reset
         if self.get_access() == "RW":
-            self._field_mirrored = wr_val
+            self._field_mirrored = wr_val & _mask
         if self.get_access() == "RC":
             self._field_mirrored = self._field_mirrored
         if self.get_access() == "RS":
@@ -396,7 +396,7 @@ class uvm_reg_field(uvm_object):
         if self.get_access() == "W0CRS":
             self._field_mirrored = self._field_mirrored & wr_val
         if self.get_access() == "WO":
-            self._field_mirrored = wr_val
+            self._field_mirrored = wr_val & _mask
         if self.get_access() == "WOC":
             self._field_mirrored = 0
         if self.get_access() == "WOS":
--- a/pyuvm/s20_uvm_reg.py
+++ b/pyuvm/s20_uvm_reg.py
@@ -165,7 +165,7 @@ class uvm_reg(uvm_object):
     def predict(self, value: int, direction: access_e):
         for f in self.get_fields():
             f.field_predict((
-                value >> f.get_lsb_pos()), path_t.FRONTDOOR, direction)
+                value >> f.get_lsb_pos() & ((1 << f.get_n_bits())-1)), path_t.FRONTDOOR, direction)
 
     # set prediction from parent to fields
     def set_prediction(self, pred_type: predict_t):
--- a/tests/pytests/test_uvm_ral.py
+++ b/tests/pytests/test_uvm_ral.py
@@ -135,7 +135,7 @@ def test_simple_reg_model():
 
     LSR.reset()
     assert LSR.get_mirrored_value() == 0
-    LSR.predict(32, access_e.UVM_WRITE)
-    assert LSR.get_mirrored_value() == 32
+    LSR.predict(12, access_e.UVM_WRITE)
+    assert LSR.get_mirrored_value() == 12
     for field in LSR.get_fields():
         print(field.get_value())
--- a/tests/pytests/test_uvm_reg_block.py
+++ b/tests/pytests/test_uvm_reg_block.py
@@ -285,7 +285,7 @@ def test_simple_reg_model():
 
     LSR.reset()
     assert LSR.get_mirrored_value() == 0
-    LSR.predict(32, access_e.UVM_WRITE)
-    assert LSR.get_mirrored_value() == 32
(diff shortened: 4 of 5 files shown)


raysalemi:
@EngRaff92 , what do you think of this PR?  Does it match the IEEE spec?

raysalemi:
@hurisson , this PR fails the QA checks.

EngRaff92:
Hey @raysalemi I quite doubt the proposal change is correct. At least at first glance. The predict on write should always update the RO field with the reset value. I will keep checking and come back to this. Anyway the code proposed looks odd and could be wrong to me. Regards

raysalemi:
Thanks! 

I spent my summer preparing a physics class and never got to look at RAL.

hurisson:
@EngRaff92  
if you look at my example, then in reg.get_mirrored in the first updt_v (from the first field ARM[0]) we get the value 0xFFFF_FFFF, this is wrong. The second time from the RESERVED field [RO] we really get 0x0, but when adding (self.mirrored | updt_v) we get 0xFFFF_FFFF.

EngRaff92:
Hey @hurisson i absolutely get your point. Anyway if the bug exists must be somewhere else. The predict I based on the policy access. If read only and the direction is write then the mirrored is updated with and only with the reset value. So could you check what is the reset value of your register? Based on the reg field implementation you can either call the predict on read or the predict on write essentially. I will make a test for it in case. But I disagree on having that predict implementation anyway. Will get back to you for sure. Tanks for the concern

EngRaff92:
Alternatively is your concern on the get mirrored value instead of? Cannot really get by reading the first comment. Regards

hurisson:
> Hey @hurisson i absolutely get your point. Anyway if the bug exists must be somewhere else. The predict I based on the policy access. If read only and the direction is write then the mirrored is updated with and only with the reset value. So could you check what is the reset value of your register? Based on the reg field implementation you can either call the predict on read or the predict on write essentially. I will make a test for it in case. But I disagree on having that predict implementation anyway. Will get back to you for sure. Tanks for the concern

Hello! The reset value for my RO fields is 0x0, but I get a different value.
You can see this if you make a register with two fields [31:1] with RO access and [0] with RW, and write 0xFFFF_FFFF into this register

hurisson:
The tests fail on CI, since the prediction in the registers began to work correctly. In the current test examples (shown below) - the LSR register has only 4 fields with a size of 1 bit. In the prediction, the value 32 is written to it, and in the assertion, 32 is also expected. This is incorrect, since 32 is 'b00100000. The unit is in the sixth position, it is not written anywhere, since there is no corresponding field

```
import itertools
from pyuvm import *

class LineStatusRegister(uvm_reg):
    def __init__(self, name="LineStatusRegister", reg_width=32):
        super().__init__(name, reg_width)
        self.DR = uvm_reg_field('DR')
        self.OE = uvm_reg_field('OE')
        self.PE = uvm_reg_field('PE')
        self.FE = uvm_reg_field('FE')

    def build(self):
        self.DR.configure(self, 1, 0, 'RW', 1, 0)
        self.OE.configure(self, 1, 1, 'RW', 1, 0)
        self.PE.configure(self, 1, 2, 'RW', 1, 0)
        self.FE.configure(self, 1, 3, 'RW', 1, 0)
        self._set_lock()
        self.set_prediction(predict_t.PREDICT_DIRECT)

class Regs(uvm_reg_block):
    def __init__(self, name):
        super().__init__(name)
        self.map = uvm_reg_map('map')
        self.map.configure(self, 0)
        self.LSR = LineStatusRegister('LSR')
        self.LSR.configure(self, "0x1014", "")
        self.map.add_reg(self.LSR, "0x0")

regs = Regs('regs')
LSR = regs.LSR

LSR.reset()
assert LSR.get_mirrored_value() == 0
LSR.predict(32, access_e.UVM_WRITE)
assert LSR.get_mirrored_value() == 32
for field in LSR.get_fields():
    print(field.get_value())
```
```
Traceback (most recent call last):
  File "/workspaces/cocotb_tutorial/pyuvm_test/uvm_reg_field_error.py", line 35, in <module>
    assert LSR.get_mirrored_value() == 32
AssertionError
```



If we increase the bit depth of the last FE field -  self.FE.configure(self, **4**, 3, 'RW', 1, 0) , we will get the correct value LSR.get_mirrored_value() = 32

My second commit: 

I reviewed my edit and decided to correct the data dimension in uvm_reg.predict. The size of the data sent to the predictor in uvm_reg_field must match the field dimension:
image


Same as in SV - uvm_reg.svh - function do_predict()
image

hurisson:
After change assert value https://github.com/pyuvm/pyuvm/pull/225/commits/82b59e2f3ca1559d8e44b2173bf6e00bf584065d
Ci test passes

raysalemi:
Great. I just need @EngRaff92 to agree that this solution is part of the spec

EngRaff92:
i will keep review the code since I wanna make sure the usage of the following code will not make hassles in future:
```python
    def predict_based_on_write(self, wr_val):
        # Define an all 1 values
        _mask = int("".join(["1"] * self._size), 2)
```

thanks to all for the support

hurisson:
Hello @EngRaff92 , @raysalemi . It seems my MR can be added, my edits do not affect the above code https://github.com/pyuvm/pyuvm/pull/225#issuecomment-2335129665

raysalemi:
I am merging this Pull Request based on the conversation.

PR #233, opened 2024-12-15
Await task directly, instead of await'ing Join

### LINKED ISSUE #232: FutureWarning: `await`ing a Join trigger will return the Join trigger and not the result of the joined Task in 2.0
When I run TinyALU example, I get thos warning:
```
random and max forked
/usr/lib/python3/dist-packages/cocotb/triggers.py:886: FutureWarning:
`await`ing a Join trigger will return the Join trigger and not the
result of the joined Task in 2.0.
  ret = outcomes.Value(await trigger)
```

aelmahmoudy:
as this will be deprecated in cocotb 2.0, see cocotb commit 579bfa682cf9ec76646654fc8c7544d812c0cfc1

Fixes #232

### DIFF
--- a/examples/TinyALU/testbench.py
+++ b/examples/TinyALU/testbench.py
@@ -1,4 +1,4 @@
-from cocotb.triggers import Join, Combine
+from cocotb.triggers import Combine
 from pyuvm import *
 import random
 import cocotb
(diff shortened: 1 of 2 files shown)


raysalemi:
This is great!  I can see why they moved to Combine()

PR #242, opened 2025-03-24
Updated register base classes to use the uvm_object name properies and name methods instead of duplicates.

FrRacz:
I ran into these problems when trying to read values from registers and logging the field names with the values. 
## `uvm_reg_field`
`uvm_reg_field` uses `self._name` as a duplicate of `uvm_object`'s `._obj_name`. This is changed to use `self.get_name()` method that every `uvm_object` has.
The method `self.get_full_name()` in `uvm_field_reg` called `self._parent` which produces the following error:
```bash
File "~/.venv/lib/python3.12/site-packages/pyuvm/s19_uvm_reg_field.py", line 125, in get_full_name
                                                            return self._parent + "." + self._name
                                                                   ~~~~~~~~~~~~~^~~~~
                                                        TypeError: unsupported operand type(s) for +: 'cl_dprio_reg' and 'str'
```
To fix this the `get_full_name()` method of the parent is called and the `get_name` method of the field.
## `uvm_reg_map`
`uvm_reg_map` also had a duplicate property `.name`, this was removed and the `get_full_name` method was updated.
## `uvm_reg`
`uvm_reg` also had a duplicate protpery `._name` this was removed and calls replaced with the `.get_name()` method

### DIFF
--- a/pyuvm/s19_uvm_reg_field.py
+++ b/pyuvm/s19_uvm_reg_field.py
@@ -80,7 +80,7 @@ class uvm_reg_field(uvm_object):
         # check the reset value is not beyond the MAX
         if (self._reset >= (2**self._size)):
             uvm_error(self._header, f"Reset value for REG \
-                      feild : {self._name} is [{self._reset}] \
+                      feild : {self.get_name()} is [{self._reset}] \
                       is beyond the MAX valoue given \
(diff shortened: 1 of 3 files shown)


raysalemi:
Thank you for fixing this!

PR #243, opened 2025-03-31
Implement pre_body and post_body for sequences

clara-h-b:
An improvement of the uvm_sequences by adding the `pre_body()` and `post_body()` functions, the implementation is based on the implementation in UVM.  As I have found that I'm missing the functionality given by the pre and post body of the sequences. Especially, when creating larger testbenches and multiple sequences to reduce the amount of repeated code across sequences. 

# Implementation
The implementation will automatically invoke the `pre_body()` before the `body()` and the `post_body()` after it. The implementation does not affect the behavior of sequences as is unless they are extended by the user.  The implementation has been done as close to the original UVM implementation as possible, see code below. 

The original UVM implementation :
```
if (call_pre_post == 1) begin
    m_sequence_state = UVM_PRE_BODY;
    #0;
    pre_body();
end

[...] body(); [...]

if (call_pre_post == 1) begin
    m_sequence_state = UVM_POST_BODY;
    #0;
    post_body();
end
```

The adapted pyuvm implementation:
```
        if call_pre_post:
            await self.pre_body()
        await self.body()
        if call_pre_post:
            await self.post_body()
```

Furthermore, I have created a test testing the implementation in `test_14_15_python_sequences.py`

### DIFF
--- a/pyuvm/s06_reporting_classes.py
+++ b/pyuvm/s06_reporting_classes.py
@@ -52,6 +52,7 @@ class PyuvmFormatter(SimColourLogFormatter):
         :param record: The log record
 
         """
+        msg_temp = record.msg
         new_msg = f"[{self.full_name}]: {record.msg}"
         record.msg = new_msg
         name_temp = record.name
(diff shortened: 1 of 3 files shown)


raysalemi:
This PR fails the flake8 style tests. It has lines that are too long.

raysalemi:
Thank you for adding this functionality!

PR #251, opened 2025-06-30
uvm_reg_block methods fix

HaMeNopeDiot:
Changes:
- The method `_get_registers()` has been renamed to `get_registers()`, as it is named in the specification "IEEE Standard for Universal Verification Methodology Language Reference Manual (Revision of IEEE Std 1800.2-2017)". A comment with the section number for this method has been added. All usages of that method were fixed. It seems that `_get_registers()` was previously called `blk_get_registers()`, so it was also renamed to `get_registers()`.

- The method `_blk_get_fields()` has been renamed to `get_fields()`, as it is named in the UVM Specification. A comment with the section number for this method has been added.

- Previously, the methods `get_fields()` and `get_registers()` returned a nested list. This has been fixed, and now the methods return flat lists, as specified in UVM.

- Added the `hier` argument to both methods and rewrote some logic for it, since it exists in the UVM Specification method definitions (sections 18.1.3.7 and 18.1.3.8).

- Added tests for `get_fields()` and `get_registers()`:
   - `test_reg_block_with_sub_blocks()` - tests `get_field()` in a reg block with sub reg block (checks that the hier argument works).
   - `test_reg_block_get_field_empty_reg()` - tests `get_field()` when the reg block is empty.
   - `test_reg_block_get_field_single_reg()` - tests `get_field()` with the reg block contains one register.
   - `test_reg_block_get_field_multiple_regs()` - tests `get_field()` with the reg block contains more than one register.
   - `test_reg_block_get_field_sub_reg_block()` - tests `get_field()` with the reg block contains a sub reg block (checks that the hier argument works).

### DIFF
--- a/pyuvm/s18_uvm_reg_block.py
+++ b/pyuvm/s18_uvm_reg_block.py
@@ -167,8 +167,9 @@ class uvm_reg_block(uvm_object):
                                        be locked"))
         return local_reg_collector
 
-    # blk_get_fields
-    def blk_get_fields(self) -> list:
+    # get_fields
+    # 18.1.3.8
+    def get_fields(self) -> list:
         local_field_collector = []
         for r in self.get_registers():
             local_field_collector.append(r.get_fields())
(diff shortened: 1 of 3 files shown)


raysalemi:
Thank you for your hard work @HaMeNopeDiot !

PR #253, opened 2025-07-10
Method predict() fix

HaMeNopeDiot:
## Changes

- **Added `uvm_reg_byte_en_t` type**  
  - Based on IEEE Standard for Universal Verification Methodology Language Reference Manual (Revision of IEEE Std 1800.2-2017) — section 17.2.1.5.

- **Fixed `status` field in `uvm_reg_item` class**  
  - Updated the field type to match the UVM specification — section 19.1.1.2.6.
  - Fixed the `set_status()` method in the validity check section.

- **Rewrote `get_value()` and `set_value()` methods in `uvm_reg_item`**  
  - Updated logic to conform with UVM specification — section 19.1.1.2.4.  
  - Also fixed incorrect handling of field value length in `get_value_size()`.

- **Added fields `fname` and `lineno` in some classes**  
  - Fields added to `uvm_reg`, `uvm_reg_field`, and `uvm_reg_item`.
  - In `uvm_reg_item`, added corresponding getter and setter methods as per UVM specification — sections 19.1.1.2.13-14.

- **Rewrote `predict()` logic in `uvm_reg` and `uvm_reg_field` classes**  
  Updated to align with UVM specification:
  - **In `uvm_reg_field` (section 18.5.5.17):**
    - Renamed `predict_based_on_write()` — `predict_write()`
    - Renamed `predict_based_on_read()` — `predict_read()`
    - Rewrote logic in these methods to be used by `predict()`
    - Introduced `do_predict()` as the main body of `predict()`, replacing removed `field_predict()`
    - Removed deprecated `prediction` field and its related methods
    - Fixed calculation of `_mask` in several functions
  - **In `uvm_reg` (section 18.4.4.15):**
    - Removed deprecated `set_prediction()` method

- **Updated tests**
  - Removed calls to deleted methods.
  - Fixed argument types in `predict()` usage.
  - Main test logic remains unchanged.

- **Added new tests for `predict()` functionality**
  - `test_reg_simple_predict()` – Tests registers with basic access types (RW/RO/WO).
  - `test_reg_predict_edge_cases()` – Tests edge cases in register prediction.
  - `test_reg_predict_with_spaces()` – Tests register prediction when fields are separated by empty space (reserved bits).
  - Added helper functions:
    - `make_expected_value()` – Builds expected values for predict assertions.
    - `predict_reg()` – Utility to test register prediction with different prediction types.

- **Fixed pytest warnings**  
  - Resolved "PytestUnknownMarkWarning: Unknown pytest.mark by adding a `pytest.ini` file with all required test marks.

- **Minor code adjustments**
  - Updated imports where necessary.
  - Adjusted whitespace and formatting to satisfy linters.

## Note

The `predict()` method in `uvm_reg_field` is not the final, specification-compliant version, but it is a working implementation. Its drawbacks include the lack of a check for parent register occupancy when performing a `UVM_PREDICT_DIRECT` prediction, as well as no support for the `uvm_reg_cbs` class, which has not yet been implemented in PyUVM.

Also the `predict()` method now has a slightly different argument sequence in order to comply with the UVM specification. Therefore, calls to `predict()` in other projects will need to be fixed in order to use it correctly.

The `set_prediction()` method is no longer needed, as `predict()` — according to the specification — already includes prediction mode handling. Thus, its usage should also be cleaned up in other projects.

It's quite possible that some testbenches relying on the previous behaviour of `predict()` may break. However, in these commits, I have fixed `predict()` to make it specification-compliant.


raysalemi:
Wow! Fantastic work!

PR #257, opened 2025-08-19
Cocotb 2.0.0b1 adoption

### LINKED ISSUE #256: Requesta: Make compatible with Cocotb 2.0.0
# Intro
Seems like a natural step to update to be compatible with upcomming Cocotb v2.0?

sibeov:
# Intro
My first pull-request to a open-source project ever. Hope I can get some pointers and perhaps someone to review some of the
stuff I've done?

## Main changes
- Updated the `pyproject.toml` file (converted the `setup.py` to a `pyproject`)
- Added a `poetry.lock` file (Poetry uses the `pyproject.toml` file directly)
- Added a `flake.nix` and `flake.lock` file for those who uses NixOS / Nix
- Updated dependencies in `pyproject.toml` to `cocotb 2.0.0b1` and `pytest`
- Bumped the required python version to `3.9`. Should consider jumping to `3.13`. I cannot see a reason not to? Yes yes. Backward compatability etc. How about we give each other a push to adopt new tech asap?

## TODOs?
- Perhaps a review?
- Maybe someone could double-check with more specific tests or something?

## Notes
I cannot see any particular reason `pyuvm`should not be compatible with `cocotb 2.0.0` for now. I'll set the PR to `Draft` for now and perhaps someone could look at it when they get the chance?

## Issues
This PR should close #256


sibeov:
Executed the tests like so (note I'm using `python 3.13`):
```shell
nix develop
pytest ./tests
```

raysalemi:
Hi,

Congratulations on your first pull request.  I'm happy that you've found **pyuvm** and are moved to contribute to it!

As for this pull request, I'm not going to merge it for a few reasons.

 * The most basic is that it does not pass the regression tests.  The regression test hangs.
 * The cocotb team is already managing the transition to version 2.0.
 * I don't understand the technology you are seeking to add (nix?).

I appreciate the way you've approached this pull request process, and especially that you've asked for advice. You are clearly willing to learn.

I recommend the following for joining an open source project. 

## Start small and at the periphery

This pull request makes significant changes to the basic framework of the project.  It adds support for additional OSs and Python versions.  It attempts to update **pyuvm** to support cocotb 2.0.

sibeov:
# Replies
@raysalemi Thanks for the feedback! Just some quick replies/questions/clarifications:

1. `The most basic is that it does not pass the regression tests. The regression test hangs.`
  - How can I run these tests? I can see now that it is hung on the test stage. I guess that is perhaps naive of me to think they would pass in the CI pipeline because the pytest(s) passed locally (except for the `scratch` thing) I have somewhat more experience with gitlab so It would take perhaps some time for me to get to know the github flavor of the `yml` setup of pipelines etc.
  
2. `The cocotb team is already managing the transition to version 2.0.`
  - Is the `cocotb` team updating `pyuvm` to work with `cocotb 2.0`? My goal here was to check if `pyuvm` could (as it is now) use `cocotb 2.0.0b1` (which it seems to do locally with me (`works on my machine`)) but If they are doing that in a fork then thats great stuff.
  
3. `I don't understand the technology you are seeking to add (nix?).`
  - `nix` is (among other) a DSL for a declarative way of building software. See:
    - Primer: https://www.youtube.com/watch?v=FJVFXsNzYZQ
    - How it works: https://nixos.org/guides/how-nix-works/
    - Wiki: https://en.wikipedia.org/wiki/Nix_(package_manager)
    - And if you want to perhaps try out the distro (its mighty fun): https://nixos.org/
    - Great explanations and examples: https://www.youtube.com/@vimjoyer/videos
  - In this particular PR I'm just adding a `flake.nix` to conveniently open a shell which has all the tools, dependencies for developing and working with `pyuvm`. It is an alternative, and in my humble opinion, somewhat more convenient way on Linux 
  systems to develop. This change is quite independent of the source-code as it only relies on the `pyproject.toml` (which it does not need to do).
  
# Further
I would love to get into the details of `pyuvm` but currently I'm still just a consumer of it and planning on using/integrating it into some personal projects. But I require Python >=  3.11 and cocotb 2.0 (when it gets out of "beta"). But I will do as @cmarqu suggests and split them up into several PRs to make it more manageble perhaps?

Perhaps someone could give me a speedrun (Discord or similar) on how the CI/CL works in github? Would be nice to chat and connect not just via asynchronous comments in PRs and commits. If not I'll figure it out somehow.

raysalemi:
@sibeov, Thanks for the detailed reply! @ktbarrett has been modifying **pyuvm** to work with **cocotb 2.0**.  Also **pyuvm** is part of the **cocotb** test suite.

raysalemi:
@sibeov , thank you for your enthusiasm about improving **pyuvm**!

ktbarrett:
cocotb uses `nox` for project automation (nothing to do with nix, despite the name), and it's basically `make` in Python with some additional functionality for handling Python environments. That's what I'd suggest for improving the current project automation.

As for changing the build system, I'd suggest sticking with setuptools for maximum compatibility with people using OS maintained packages of Python and setuptools and pip, and also being behind a firewall preventing the use of arbitrary Python packages. If modernization were desired, I would go with bumping the minimum Python and setuptools version supported and moving the package info to `pyproject.toml` so tools like uv can be used, but newer setuptools can as well.

scratch/logging_test.py
```diff
@@ -1,5 +1,5 @@
 import logging
-from cocotb.log import default_config
```
ktbarrett, line 2: This is a valid fix. Might be worth getting in separately @sibeov @raysalemi

raysalemi:
It is also a great first pull request.

Issue #264, opened 2025-09-19
KeyError when logging a uvm_factory


DedkovEA:
Seems like `utility_classes.Override`'s `__str__` function produces `KeyError` when the path of instance override is longer than 29 characters.

Inspection of the function code shows a simple error:
```python
def __str__(self):
    """
    For printing out the overrides
    :return: str
    """
    if self.type_override is not None:
        to = "Type Override: " + f"{self.type_override.__name__}"
    else:
        to = "Type Override: None"
    ss = f"{to:25}" + " || "
    if self.inst_overrides is not None:  # WARNING: see __init__ method, inst_overrides can not be None
        ss += "Instance Overrides: "
        first = True
        for inst in self.inst_overrides:
            if not first:
                ss += " | "
            first = False
            if len(inst) > 29:
                inst = inst[:29]  # Here the instance path is changed, and is no longer a valid key for inst_overrides dict 
            ss += inst + f" => {self.inst_overrides[inst].__name__}"  # ERROR: we refer to non existing key

    return ss
```

The proposition is to rewrite this function into something like
```python
def __str__(self):
    """
    For printing out the overrides
    :return: str
    """
    if self.type_override is not None:
        to = "Type Override: " + f"{self.type_override.__name__}"
    else:
        to = "Type Override: None"
    ss = f"{to:25}" + " || "
    inst_overrides_str = " | ".join([
        f"{inst_path[:29] if len(inst_path) > 29 else inst_path} => {inst_override.__name__}" 
        for inst_path, inst_override in self.inst_overrides.items()
    ])
    ss += f"Instance Overrides: {inst_overrides_str}" if inst_overrides_str else ""
   
    return ss
```

raysalemi:
Could you create a PR with a test?

DedkovEA:
I have no experience in contributing, but I'll try my best

raysalemi:
This is a great starter issue to dip your toe into the water of open source development.  

You can run the regression by installing `tox` (`% pip install tox`) and typing `tox` at the top level of your cloned repo.

DedkovEA:
Closed with #265

PR #265, opened 2025-09-20
Fix of Override's __str__() method

DedkovEA:
This PR is a fix of #264.

- New test of `str(uvm_factory)` method was added, which aim is to test correct logging of instance overrides with long paths (it is assumed, that at least first 24 characters of the path should be printed, which is true for the previous logic, but is up to discussion).
- Fixed incorrect logic for `Override.__str__` method in these cases.

### DIFF
--- a/pyuvm/utility_classes.py
+++ b/pyuvm/utility_classes.py
@@ -79,16 +79,12 @@ class Override:
         else:
             to = "Type Override: None"
         ss = f"{to:25}" + " || "
-        if self.inst_overrides is not None:
-            ss += "Instance Overrides: "
-            first = True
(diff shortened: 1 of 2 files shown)


raysalemi:
I think this code has an error.  Why is 

`if self.inst_overrides is not None:` removed on line 82?

raysalemi (approved):
Ah!  Nicely done.

raysalemi:
Thank you for improving **pyuvm**!

DedkovEA:
Thank you for such wonderful framework! 
Appreciate your amiability a lot. Do I need to take some actions to close my issue?

PR #268, opened 2025-10-01
Update CI

ktbarrett:
Updates the CI to test all versions of Python and all versions of cocotb supported to prevent regressions. Also changes the VHDL simulator from GHDL to NVC for better cocotb support.


ktbarrett:
The CI update looks good now, there's just a test error with cocotb 1.6 that should be easy to resolve, and a failure with GHDL on Python 3.6 that might be harder to diagnose.

PR #280, opened 2025-10-03
Do 4.0.0 release

ktbarrett:
I also added a version field so that users can distinguish versions. This is common in Python libraries. In the future we can introduce ways to automatically generate this file.

Will make tag after merge.


raysalemi (approved):
These are great!  You've done a fantastic job, Kaleb!  Thank you so much!

PR #294, opened 2025-10-17
Support inheriting from decorated test again

### LINKED ISSUE #293: @pyuvm.test() decorator breaks class inheritance with cocotb 2.0+
## Description

The `@pyuvm.test()` decorator in `pyuvm/extension_classes.py` breaks class inheritance when using cocotb 2.0+ due to a change in how cocotb's `@cocotb.test()` decorator works.

## Problem

In cocotb 2.0+, the `@cocotb.test()` decorator returns a `Test` class instance instead of the decorated function (as it did in cocotb < 2.0). When `@pyuvm.test()` returns this `Test` instance, it breaks Python class inheritance because:

1. The parent class becomes a `Test` instance instead of the original class
2. When a child class tries to inherit from it, Python attempts to call `Test.__init__()` with arguments intended for the original class  
3. `Test.__init__()` doesn't accept these arguments, causing a `TypeError`

## Example that fails

```python
@pyuvm.test()
class BaseTest(uvm_test):
    # base test implementation
    def __init__(self, name, parent, other_things):
        ...

@pyuvm.test() 
class ExtendedTest(BaseTest):  # This fails with cocotb 2.0+
    # extended test implementation
    def build_phase(self):
        super().build_phase()
        ...

    def connect_phase(self):
        super().connect_phase()
        ...

```

## Error message

```
TypeError: Test.__init__() takes 1 positional argument but 4 were given
```

## Root cause

In `extension_classes.py`, the decorator always returns `test_obj`, which in cocotb 2.0+ is a `Test` class instance rather than the original class.

## Proposed fix

Modify the `test()` function in `pyuvm/extension_classes.py` to detect the cocotb version and handle both cases:

```python
def test(
    timeout_time=None,
    timeout_unit="step",
    expect_fail=False,
    expect_error=(),
    skip=False,
    stage=None,
    keep_singletons=False,
    keep_set=set(),
):
    if cocotb_version_info >= (1, 7, 0) and stage is None:
        stage = 0

    def decorator(cls):
        # create cocotb.test object to be picked up RegressionManager
        @cocotb.test(
            timeout_time=timeout_time,
            timeout_unit=timeout_unit,
            expect_fail=expect_fail,
            expect_error=expect_error,
            skip=skip,
            stage=stage,
        )
        @functools.wraps(cls)
        async def test_obj(_):
            await uvm_root().run_test(
                cls, keep_singletons=keep_singletons, keep_set=keep_set
            )

        if cocotb_version_info >= (2, 0):
            # Get the module where the class is defined
            import sys
            frame = sys._getframe(1)  # Get caller's frame
            caller_globals = frame.f_globals
            
            # Create a unique name for the test object in the module's namespace
            # This ensures cocotb's test discovery can find it
            test_obj_name = f"_pyuvm_test_{cls.__name__}_{id(cls)}"
            caller_globals[test_obj_name] = test_obj
            
            # Return the original class to preserve inheritance
            return cls
        else:
            # For cocotb < 2.0, return the test object as before (backward compatibility)
            return test_obj

    return decorator
```

## Environment

- **pyuvm version:** 4.0.0+
- **cocotb version:** 2.0+
- **Python version:** 3.12
- **Operating System:** Linux

## Backward compatibility

The proposed fix maintains full backward compatibility with cocotb < 2.0.

ktbarrett:
Closes #293.

### DIFF
--- a/examples/TinyALU/testbench.py
+++ b/examples/TinyALU/testbench.py
@@ -262,8 +262,9 @@ class AluEnv(uvm_env):
         self.driver.ap.connect(self.scoreboard.result_export)
 
 
-class AluTestBase(uvm_test):
-    """Base class for ALU tests with random and max values"""
+@pyuvm.test()
+class AluTest(uvm_test):
+    """Test ALU with random and max values"""
 
     def build_phase(self):
         self.env = AluEnv("env", self)
@@ -278,12 +279,7 @@ class AluTestBase(uvm_test):
 
 
 @pyuvm.test()
-class AluTest(AluTestBase):
-    """Test ALU with random and max values"""
-
-
-@pyuvm.test()
-class ParallelTest(AluTestBase):
+class ParallelTest(AluTest):
     """Test ALU random and max forked"""
 
     def build_phase(self):
@@ -292,7 +288,7 @@ class ParallelTest(AluTestBase):
 
 
 @pyuvm.test()
-class FibonacciTest(AluTestBase):
+class FibonacciTest(AluTest):
     """Run Fibonacci program"""
 
     def build_phase(self):
@@ -302,7 +298,7 @@ class FibonacciTest(AluTestBase):
 
 
 @pyuvm.test(expect_fail=True)
-class AluTestErrors(AluTestBase):
+class AluTestErrors(AluTest):
     """Test ALU with errors on all operations"""
 
     def start_of_simulation_phase(self):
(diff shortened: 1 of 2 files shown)


raysalemi:
Hi @ktbarrett ,

My top priority is that whatever we do be compatible with the examples in the book.  

I'm open to extensions that don't change the basic behavior.  

What if we have a different decorator name for more complex behavior?

Issue #302, opened 2025-11-05
Make autodoc2 respect `__all__`


ktbarrett:
This requires a bit of configuration (https://sphinx-autodoc2.readthedocs.io/en/latest/quickstart.html#documenting-only-the-public-api-via-all) and also adding `__all__` to many existing modules.

The alternative is we trust the module author was respectful for Python conventions surrounding private attributes and re-exporting and remove `__all__` from the other modules.

PR #288, opened 2025-10-08
Add all definitions of UVM-RAL classes

RasmusGOlsen:
This commit marks the beginning of the cleanup process and updates the UVM Register Layer to align with IEEE 1800.2-2020 standards. I have created a new directory called **reg** where all UVM Register Layer files are now located. The structure follows the original SystemVerilog implementation found [here](https://github.com/accellera-official/uvm-core) to facilitate tracking progress and comparisons. 

Below is a list of tasks that will help with the transition.

- [x] Define all uvm_reg* classes
- [x] Define methods for all uvm_reg* classes (in progress)
- [x] Migrate existing implementation
    - [x] uvm_reg_field
    - [x] uvm_reg
    - [x] uvm_mem
    - [x] uvm_reg_block
    - [x] uvm_reg_map
    - [x] uvm_reg_item
    - [x] uvm_reg_adapter
    - [x] uvm_reg_predict
- [x] Create test cases for each class

Resolves: #226 , Resolves: #227

PR #318, opened 2025-12-13
Make implementation private and properly re-export symbols

### LINKED ISSUE #303: Make all modules beyond the top-level `pyuvm` private
i.e. `pyuvm.s05_base_classes` -> `pyuvm._s05_base_classes`.

ktbarrett:
Closes #303. Assists with #302.

* Made all implementation modules private
* Replaced `import *` usage in `__init__.py` with explicit list of imports to prevent transitively importing implementation functions/classes
* Properly re-exported symbols in `__init__.py` by fixing up object `__module__`

Most of these refactorings were accomplished with VSCode's Python-aware file renaming functionality and Copilot queries to do the rather tedious re-exporting properly.

The only manual thing I did that may be contentious is I removed the seemingly deliberate re-exporting of the default logging levels: `CRITICAL`, `ERROR`, `WARNING`, `INFO`, `DEBUG` and `NOTSET`.

### DIFF
--- a/examples/TinyALU/testbench.py
+++ b/examples/TinyALU/testbench.py
@@ -1,3 +1,4 @@
+import logging
 import random
 
 # All testbenches use tinyalu_utils, so store it in a central
@@ -136,7 +137,7 @@ class FibonacciSeq(uvm_sequence):
             prev_num = cur_num
             cur_num = sum
         uvm_root().logger.info("Fibonacci Sequence: " + str(fib_list))
-        uvm_root().set_logging_level_hier(CRITICAL)
+        uvm_root().set_logging_level_hier(logging.CRITICAL)
 
 
 class Driver(uvm_driver):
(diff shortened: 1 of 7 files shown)


ktbarrett:
@raysalemi Do you have any opinions on this PR? pyuvm continues to work as detailed in the book, since the book uses `from pyuvm import *`. The only question I had was on the inclusion of the Python logging levels: `pyuvm.INFO`, `pyuvm.DEBUG`, etc.

Making things private means we can move around modules like was done in #288 without breaking the API.

Specifying `__all__` means that users don't implicitly import random implementation modules, potentially overriding similarly named imports when using `from pyuvm import *` like currently occurs. And will also make the autodoc2-based documentation not include random implementation details either.

raysalemi:
As long as the examples in the book work as shown in the book, I don't see a problem.

PR #333, opened 2026-02-13
Implement Core UVM Callback Infrastructure

RasmusGOlsen:
This PR introduces the base infrastructure for the UVM Callback mechanism. In SystemVerilog, callbacks require heavy macro usage (uvm_register_cb, uvm_do_callbacks) to manage type-safe queues. This implementation leverages Python’s dynamic nature to provide a more streamlined experience.

**Key Changes:**
* uvm_callback: Base class for user-defined callback methods.
* uvm_callbacks: The management interface for adding/deleting callbacks.
* uvm_callback_iter: Iterator for traversing typed callback queues.
* uvm_do_callbacks: A functional equivalent to the SV macro to execute hooks.

**Implementation Notes**
* **Single Singleton Architecture:** Unlike SystemVerilog UVM, which creates a uvm_callbacks singleton per type via parameterization, this implementation uses a single global singleton for all callbacks. It uses a key (object/type, callback_type) to maintain the necessary separation of concerns.

* **Redundancy of uvm_register_cb:** Since Python handles dynamic attribute assignment, we do not require an explicit registration step. The first time a callback is added to a class type or instance, the registry is initialized automatically.

* **Simplified add():** In SystemVerilog, one must specify `uvm_callbacks#(T, CB)::add(null, cb)`. In this implementation, providing the class type directly to the uvm_callbacks constructor handles global (type-wide) callbacks.

* **Naming Limitations:** add_by_name and delete_by_name are implemented for API parity but are currently non-functional as they depend on `uvm_root.find_all()`, which is not yet present in the core.

### DIFF
--- a/tests/pytests/test_uvm_callback.py
+++ b/tests/pytests/test_uvm_callback.py
@@ -1,0 +1,458 @@
+import pytest
+
+from pyuvm import uvm_callback, uvm_object
+
+pytestmark = pytest.mark.usefixtures("initialize_pyuvm")
+
+
+class test_callback(uvm_callback):
+    """Test callback class extending uvm_callback"""
+
+    def __init__(self, name: str = "test_callback"):
+        super().__init__(name)
+        self.call_count = 0
+        self.last_args = None
+        self.last_kwargs = None
+
+    def on_event(self, *args, **kwargs):
+        """Test callback method"""
+        self.call_count += 1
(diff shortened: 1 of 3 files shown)


raysalemi:
This is interesting!  I've never worked with callbacks.  

Let's merge it in!

PR #357, opened 2026-04-14
Remove broad exception fallback in `uvm_factory()` and type-check test_name in `run_test`

FrRacz:
Both `create_component_by_name` and `create_object_by_name` had a exception fallback to use the `requested_type_name` as the `requested_type` if the lookup in the factory classes failed.
This is not in line with the UVM standard, as is states 
>  If the factory does not recognize the requested_type_name, an error shall be generated and a null handle returned[^1]

Additionally is creates confusion as the methods essentially function as create by name or type.
Therefore is is removed and an exception is thrown instead.

The check for `requested_type_name` is extended to also catch empty strings, as the default value is an empty string.

The `run_test` function is extended to correctly call `request_component_by_type`/`request_component_by_name` depending on if a string is passed.

[^1]: IEEE Std 1800.2-2020 p. 78

### DIFF
--- a/pyuvm/_s08_factory_classes.py
+++ b/pyuvm/_s08_factory_classes.py
@@ -238,10 +238,17 @@ class uvm_factory(metaclass=Singleton):
 
         Create an object using a string to define its uvm_object type.
         """
+        if name is None or name is "":
+            raise UVMFactoryError(
+                "Parameter name must be specified in create_object_by_name"
+            )
+
         try:
             requested_type = FactoryData().classes[requested_type_name]  # noqa
         except KeyError:
-            requested_type = requested_type_name
+            raise UVMFactoryError(
+                f"Requested type {requested_type_name} not in uvm_factory()"
+            )
 
         new_obj = self.create_object_by_type(requested_type, parent_inst_path, name)
         return new_obj
@@ -289,7 +296,7 @@ class uvm_factory(metaclass=Singleton):
         :raises: UVMFactoryError if the type is not in the factory
         :return: A uvm_object with the name given
         """
-        if name is None:
+        if name is None or name is "":
             raise UVMFactoryError(
                 "Parameter name must be specified in create_component_by_name"
             )
@@ -297,7 +304,9 @@ class uvm_factory(metaclass=Singleton):
         try:
             requested_type = FactoryData().classes[requested_type_name]  # noqa
         except KeyError:
-            requested_type = requested_type_name
+            raise UVMFactoryError(
+                f"Requested type {requested_type_name} not uvm_factory()"
+            )
 
         new_obj = self.create_component_by_type(
             requested_type, parent_inst_path, name, parent
--- a/pyuvm/_s08_factory_classes.py
+++ b/pyuvm/_s08_factory_classes.py
@@ -253,7 +253,7 @@ class uvm_factory(metaclass=Singleton):
             else:
                 raise UVMFactoryError(
                     f"Requested type {requested_type_name} not in "
-                     "uvm_factory()  or overrides"
+                    "uvm_factory()  or overrides"
                 )
 
         new_obj = self.create_object_by_type(requested_type, parent_inst_path, name)
@@ -310,7 +310,7 @@ class uvm_factory(metaclass=Singleton):
             else:
                 raise UVMFactoryError(
                     f"Requested type {requested_type_name} not in "
-                     "uvm_factory()  or overrides"
+                    "uvm_factory()  or overrides"
                 )
 
         new_obj = self.create_component_by_type(
--- a/pyuvm/_s08_factory_classes.py
+++ b/pyuvm/_s08_factory_classes.py
@@ -243,12 +243,18 @@ class uvm_factory(metaclass=Singleton):
                 "Parameter name must be specified in create_object_by_name"
             )
 
+        # Find the requested_type_name in registered classes or overrides if
+        # the string has been registered with an override.
         try:
             requested_type = FactoryData().classes[requested_type_name]  # noqa
         except KeyError:
-            raise UVMFactoryError(
-                f"Requested type {requested_type_name} not in uvm_factory()"
-            )
+            if requested_type_name in self.fd.overrides:
+                requested_type = requested_type_name
+            else:
+                raise UVMFactoryError(
+                    f"Requested type {requested_type_name} not in "
+                     "uvm_factory()  or overrides"
+                )
 
         new_obj = self.create_object_by_type(requested_type, parent_inst_path, name)
         return new_obj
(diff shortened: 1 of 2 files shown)


ktbarrett:
pre-commit.ci autofix

pyuvm/_s08_factory_classes.py
```diff
@@ -238,10 +238,17 @@ def create_object_by_name(self, requested_type_name, parent_inst_path="", name="
 
         Create an object using a string to define its uvm_object type.
         """
+        if name is None or name is "":
```
ktbarrett:
```suggestion
        if name is None or name == "":
```

pyuvm/_s08_factory_classes.py
```diff
@@ -289,15 +296,17 @@ def create_component_by_name(
         :raises: UVMFactoryError if the type is not in the factory
         :return: A uvm_object with the name given
         """
-        if name is None:
+        if name is None or name is "":
```
ktbarrett:
```suggestion
        if name is None or name == "":
```

pyuvm/_s13_uvm_component.py
```diff
@@ -497,9 +497,15 @@ async def run_test(self, test_name, keep_singletons=False, keep_set=set()):
             factory.clear_overrides()
         self.clear_children()
         ObjectionHandler().clear()
-        self.uvm_test_top = factory.create_component_by_name(
-            test_name, "", "uvm_test_top", self
-        )
+        if type(test_name) is str:
```
ktbarrett:
```suggestion
        if isinstance(test_name, str):
```

ktbarrett:
I'm no UVM expert, but I have a copy of IEEE 1800.2 and an LLM.

> ● I read both commits on abc/master (ef3fa04 and 214e81b). The patch is mostly aligned with what we discussed, but I see two issues worth flagging before this goes in.
> 
>   What the patch does
> 
>   _s08_factory_classes.py:
>   - create_object_by_name / create_component_by_name: removes the except KeyError: requested_type = requested_type_name fallback and raises UVMFactoryError instead.
>   - Both methods also reject name is None or name == "" (after the 214e81b fix for is "" → ==).
> 
>   _s13_uvm_component.py:
>   - run_test branches on isinstance(test_name, str): string → create_component_by_name, otherwise → create_component_by_type. This is necessary because before the s08 change, passing a class to create_component_by_name happened to work
>   (KeyError → string fallback → __find_override accepted the class anyway); after, it would raise.
> 
>   Issue 1 — the patch breaks a spec-required use case
> 
>   Look at §8.3.1.4.1 (page 77, the override-by-name text I pulled earlier):
> 
>   ▎ "When overriding by name, the original_type_name typically refers to a preregistered type in the factory. It may, however, be any arbitrary string. Future calls to any of the create_* methods with the same string and matching instance path
>   ▎  produce the type represented by override_type_name..."
> 
>   So the spec mandates this flow:
>   factory.set_inst_override_by_name("MyAbstractFoo", "ConcreteFoo", "env.*")
>   obj = factory.create_object_by_name("MyAbstractFoo", "env.x")  # must return ConcreteFoo
> 
>   "MyAbstractFoo" is never a registered class — it only exists as an override key. The old except KeyError: requested_type = requested_type_name fallback was specifically what let this work: the string was passed through to __find_override,
>   which looked it up in self.fd.overrides.
> 
>   After the patch, that call raises UVMFactoryError("Requested type MyAbstractFoo not in uvm_factory()") even though there's a perfectly good override registered. The spec's "error+null on unrecognized name" rule is supposed to coexist with
>   the "arbitrary-string override" rule — the patch enforces the first and forgets the second.
> 
>   The right shape is: try the class lookup; on miss, only raise if the string also isn't a known override key. Something like:
>   try:
>       requested_type = FactoryData().classes[requested_type_name]
>   except KeyError:
>       if requested_type_name in self.fd.overrides:
>           requested_type = requested_type_name  # let __find_override resolve it
>       else:
>           raise UVMFactoryError(
>               f"Requested type {requested_type_name!r} is not registered "
>               f"and has no override"
>           )
> 
>   (You may want a test that exercises set_inst_override_by_name / set_type_override_by_name with an arbitrary, unregistered original name and then creates by that name — I suspect there isn't one, otherwise the patch would have failed CI.)
> 
>   Issue 2 — empty-name check is wrong for create_object_by_name
> 
>   The spec signature is:
>   create_object_by_name(
>     string requested_type_name,
>     string parent_inst_path = "",
>     string name = ""        // default empty
>   )
>   and §8.3.1.5 says: "New objects use the given name, if provided." So "" is a legal, spec-default value for object creation — it means "unnamed." Rejecting it makes create_object_by_name(type_name) (the no-name overload) raise, which is a
>   regression and breaks parity with create_object_by_type which still accepts name="".
> 
>   For create_component_by_name the empty check is fine — the spec signature has no default for name and components need a name to live in the hierarchy.
> 
>   So the fix is: drop the name == "" part of the check in create_object_by_name (keep name is None if you want — though even that wasn't a check in the by-type variant). Keep the full check only in create_component_by_name.
> 
>   Issue 3 — minor, run_test type handling
> 
>   if isinstance(test_name, str): ... else: create_component_by_type(test_name, ...) will pass non-string non-class values (e.g. None) into create_component_by_type, where they'll trip the issubclass(requested_type, uvm_void) assertion in
>   __find_override with a confusing message. Not blocking — but a tighter check (isinstance(test_name, type) and issubclass(test_name, uvm_component)) and a clean TypeError would be friendlier. Up to you.
> 
>   Verdict
> 
>   The direction is right and matches what we discussed (raise instead of returning None, fix run_test to dispatch by type). Two real problems to address before merging:
> 
>   1. The arbitrary-string override path (a spec-required behavior) is broken — needs the override-key check shown above.
>   2. create_object_by_name shouldn't reject name == "" — that's the spec default.
> 
>   214e81b's `is ""` → `==` fix is good and worth keeping regardless.

raysalemi:
> I'm no UVM expert, but I have a copy of IEEE 1800.2 and an LLM.
> 
> > ● I read both commits on abc/master ([ef3fa04](https://github.com/pyuvm/pyuvm/commit/ef3fa044df6b7ebfb94a1c47622d27b2228734a5) and [214e81b](https://github.com/pyuvm/pyuvm/commit/214e81b335a6749124b5ee843d2ca4f6296c9901)). The patch is mostly aligned with what we discussed, but I see two issues worth flagging before this goes in.

> > Verdict
> > The direction is right and matches what we discussed (raise instead of returning None, fix run_test to dispatch by type). Two real problems to address before merging:
> > 
> > 1. The arbitrary-string override path (a spec-required behavior) is broken — needs the override-key check shown above.
> > 2. create_object_by_name shouldn't reject name == "" — that's the spec default.
> > 
> > [214e81b](https://github.com/pyuvm/pyuvm/commit/214e81b335a6749124b5ee843d2ca4f6296c9901)'s `is ""` → `==` fix is good and worth keeping regardless.

This is a great analysis!

If I remember correctly, "" is the way SystemVerilog says `None`.  So, that's why it responded to both.

Do we think the patch will break someone's code?  I don't see why it would.

FrRacz:
@ktbarrett Thanks for the analysis. I missed the two issues. I updated the code to include the suggested fallback for arbitrary strings and removed the check for an empty name parameter is create_object_by_name

ktbarrett:
pre-commit.ci autofix

ktbarrett (approved):

PR #366, opened 2026-06-03
Automate releases

### LINKED ISSUE #343: Add release artifact smoke testing
All commits to master should trigger a release flow which builds the sdist and wheel and tests that installation from those objects and running the tests (or some subset) on master against them.

These build artifacts can be uploaded so that users can download the sdist and wheels per commit as well.

We still need to run the development flow on these commits to ensure that the project coverage is recorded.

ktbarrett:
This add a release smoke test that runs on every master push (closes #343), adds a full release regression test, adds an automated release to PyPI on git tag pushes (closes #344), and refactors the test flows to use a common test flow.

Also `tox` isn't buying us anything anymore, so we are removing it. The Makefile and uv already do the brunt of the work.


ktbarrett:
The Python 3.7 issue is due to how pyenv sets up the environment.

ktbarrett:
Well... there's no happy medium is there? Maybe we should remove support for 3.7 to make CI happy #368.

raysalemi:
I will spend time this summer coming up to speed on all the great work
you’ve done!

ktbarrett:
@raysalemi This is good to go now.

raysalemi:
It says it cannot merge.

ktbarrett:
No problems with rebasing here, I'll do it tomorrow.

PR #376, opened 2026-06-21
docs: note that examples/ require an editable install of the cloned source (closes #363)

### LINKED ISSUE #363: TinyALU example fails when combining pip install with git clone
Follownig the README in order to run the TinyALU example, the simulation fails with a TypeError, 
 if the package is installed via pip and the example is cloned from repo.

This might be a common pitfall for new users who follow the README step by step.

## Steps to reproduce 

1. pip install pyuvm
2. git clone https://github.com/pyuvm/pyuvm
3. cd pyuvm/examples/TinyALU
4. make SIM=icarus

## Error

TypeError: uvm_component.raise_objection() takes 1 
positional argument but 2 were given

## Root Cause

commit d28405a updated testbench.py to use:
    self.raise_objection("Keep simulation alive")

This requires the description="" parameter added in the same commit. 
However, PyPI release 4.0.1 predates this commit and still has:
    def raise_objection(self):

So pip install pyuvm gives the old API, but git clone gives the new testbench — causing a mismatch.

## Suggested Fix

The "Running the simulation" section of the README 
could clarify that the examples require the cloned 
version to be installed:

    git clone https://github.com/pyuvm/pyuvm
    cd pyuvm
    pip install -e .        # ← add this line
    cd examples/TinyALU
    make SIM=icarus

Alternatively, a new PyPI release that includes d28405a would also resolve the issue.

## Environment

- pyuvm: 4.0.1 (pip)
- cocotb: 2.0.1
- Python: 3.12.3
- Simulator: Icarus Verilog 12.0
- OS: Ubuntu

seam0814:
## Summary

A user who follows the README top-down hits

\`\`\`
TypeError: uvm_component.raise_objection() takes 1 positional argument but 2 were given
\`\`\`

when they:

1. \`pip install pyuvm\` (Installation section)
2. \`git clone pyuvm\` (implied by the example path in the \"Running the simulation\" section)
3. \`cd examples/TinyALU && make sim\`

…because the cloned \`examples/testbench.py\` already uses the new
\`raise_objection(description)\` signature added in commit d28405a,
while the pinned PyPI release predates that change and still exposes
\`raise_objection(self)\` only. Imports resolve to the PyPI copy, so
the example breaks.

The \"Running from a cloned repository\" section already documents the
correct setup (\`pip install -e .\` in the cloned tree), but the
TinyALU run-through doesn't reference it, so new users land in the
broken combination by default.

Closes #363.

## Change

Single-block note at the top of the \"Running the simulation\" section:

> The example files in \`examples/\` are kept in lockstep with the latest
> pyuvm source on \`master\` and may rely on API additions that have not
> yet been published to PyPI. Install pyuvm in editable mode from your
> cloned repository (see [Running from a cloned repository] above)
> before running the example, otherwise the imports resolve to the
> older PyPI release and you may hit errors such as
> \`TypeError: uvm_component.raise_objection() takes 1 positional
> argument but 2 were given\` (see #363).

## Scope

Documentation only. Single file change (+9 lines in \`README.md\`), no
code or test changes. The reporter (@JohnTheo01) suggested this same
fix in the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,15 @@ This section demonstrates running an example simulation and then shows how the e
 
 ## Running the simulation
 
+> **Note** &mdash; The example files in `examples/` are kept in lockstep with the
+> latest pyuvm source on `master` and may rely on API additions that have not
+> yet been published to PyPI. Install pyuvm in editable mode from your cloned
+> repository (see [Running from a cloned repository](#running-from-a-cloned-repository)
+> above) before running the example, otherwise the imports resolve to the
+> older PyPI release and you may hit errors such as
+> `TypeError: uvm_component.raise_objection() takes 1 positional argument but 2 were given`
+> (see #363).
+
 The TinyALU is, as its name implies, a tiny ALU. It has four operations: ADD, AND, NOT, and MUL. This example shows us running the Verilog version of the design, but there is also a VHDL version.
 
 **cocotb** uses a Makefile to run its simulation. We see it in `examples/TinyALU`:


ktbarrett:
Dupe of #364.

seam0814:
Ah — didn't notice #364 was already open when I wrote this. Thanks for landing the fix via that one. I'll search for referenced PRs first next time before opening one.

PR #380, opened 2026-06-28
Use `uv` to source testing environment and add `uv.lock`

ktbarrett:
After #366 comes in, we will be using `uv` to source the environment for CI testing. `uv sync` without a lock file just gets the most recent versions of dependencies


raysalemi (commented):
Looks good to me!

PR #383, opened 2026-06-28
Use setuptools-git-versioning as version provider

### LINKED ISSUE #367: Post-release modernization
After 5.0 is released, we should
* Remove support for Python <3.9 (like cocotb)
* Use uv for test environment management
* Use uv.lock for repeatable tests
* Use `setuptools-git-versioning`

ktbarrett:
Closes #367

This prevents us from having to duplicate the version info in two places and them maybe accidentally getting out of sync. This tool also can automatically generate a "dev" version number by looking at the number of commits since the VERSION file was last updated. Finally, it can add a `+dirty` to the version if installing from a modified local copy, so maintainers can quickly see if the user is using a modified version when rendering assistance.

- [ ] See if CI is dirty and fix
- [ ] Consider if we need more checkout depth

I can't easily see if the version is being created appropriately, so we should get #366 in first, which changes from tox to uv+Makefile.


pyuvm/_version.py
```diff
@@ -1 +1,3 @@
-__version__ = "4.0.1"
+import importlib.metadata
+
+__version__ = importlib.metadata.version("pyuvm")
```
ktbarrett, line 3: This pulls the version information from the package's metadata instead of duplicating it in the installed package.

.git_archival.txt
```diff
@@ -0,0 +1,4 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%D$
+ref-names: $Format:%D$
```
ktbarrett, line 4: This and the change in `.gitattributes` allows the version to be filled in when running `git archive` which is what Github uses when building an archive via `Download as ZIP`. So users who do that will still get an appropriate dev version we can use to know what commit they are using.

VERSION
```diff
@@ -0,0 +1 @@
+5.0
```
ktbarrett, line 1: This file lists the next release version and should be bumped after each release.

raysalemi:
That is cool!

raysalemi (approved):

PR #393, opened 2026-07-10
Implement uvm_mem RAL core support and reg_map memory plumbing

### LINKED ISSUE #392: uvm_mem can't be instantiated (RAL section 22 memory support)
IEEE 1800.2 section 22 defines uvm_mem and uvm_mem_mam for modeling and
allocating memory through the RAL. Right now neither can be built:
uvm_mem.__init__ and uvm_mem_mam.__init__ (and the rest of
uvm_mem_mam/uvm_mem_region) raise NotImplementedError unconditionally.
This matches the README, which already marks section 22 as "Not
Implemented."

38 of uvm_mem's ~64 methods are stubs, including write, read, do_write,
do_read, burst_write, burst_read, poke, and peek, so there's no frontdoor
path either. There's also a bug in configure(): the code that creates the
mam and registers the memory with its parent block is nested inside the
branch for an invalid access value, so it never runs for a normal RW or RO
memory.

For comparison, uvm_reg has a full frontdoor implementation (async
write/read/do_write/do_read, locking via self._atomic, dispatch through
uvm_reg_item), and only stubs out backdoor, HDL paths, coverage, and
burst/peek/poke.

Commit dd30013 (RasmusGOlsen, Oct 15 2025) started on this, filling in
configure(), set_offset(), set_parent(), add_map(), and get_full_name(),
but the constructor still raises and read/write were never touched.
Nothing has changed since, and raysalemi confirmed in #261 that nobody's
working on it.

A reasonable first step would be to fix the constructor and the
configure() bug, get uvm_mem_mam/uvm_mem_region far enough to back an
address range, implement frontdoor write/read/burst_write/burst_read the
same way uvm_reg does, and add tests/pytests/reg/test_uvm_mem.py (there's
no test file for memory currently). Backdoor, HDL paths, coverage, and
poke/peek could wait for later PRs.

I'd be glad to help with this if it's useful. Does that scope look right,
or would you rather split uvm_mem core and the allocation manager into
separate PRs?

vs4848:
Adds working uvm_mem functionality per pyuvm/pyuvm#392:

- uvm_mem: fix configure()/init/_lock_model bugs that prevented memories from ever registering with their parent block; implement frontdoor write/read/burst_write/burst_read, do_write/do_read, and _check_access, mirroring the uvm_reg frontdoor pattern.
- uvm_mem: fix get_addresses() (map_info.stride -> mem_range.stride) and get_n_bytes() (operator-precedence bug that returned ~n_bits instead of the byte count).
- uvm_reg_map: implement memory address-map plumbing - mem loop in _init_address_map, add_mem(), _set_mem_offset(), get_mem_map_info(), get_mem_by_offset(), and mem branches in do_bus_write()/do_bus_read().
- uvm_reg_model: implement uvm_reg_map_addr_range.__init__.
- uvm_reg_block: add _add_memory() mirroring _add_register().
- tests: add tests/pytests/reg/test_uvm_mem.py covering name/size/ n_bits/n_bytes/access, parent/block linkage, address computation (including offset and out-of-range), map registration, and the locked-block registration-rejection path.

Backdoor access (poke/peek, do_*_backdoor) and uvm_mem_mam are left as NotImplementedError, consistent with uvm_reg's current stub state, and are deferred to a follow-up PR.

Fixes pyuvm/pyuvm#392

### DIFF
--- a/pyuvm/_reg/uvm_mem.py
+++ b/pyuvm/_reg/uvm_mem.py
@@ -538,7 +538,7 @@ class uvm_mem(uvm_object):
                 await self._do_read_frontdoor(rw, map_info)
             # TODO: post_read callbacks
             # TODO: report
-       finally:
+        finally:
             self._read_in_progress = False
 
     async def _do_read_backdoor(
--- a/pyuvm/_reg/uvm_mem.py
+++ b/pyuvm/_reg/uvm_mem.py
@@ -3,23 +3,23 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING, ClassVar
 
+from cocotb.triggers import Lock
+
 from pyuvm._error_classes import UVMFatalError
+from pyuvm._reg.uvm_reg_item import uvm_reg_item
 from pyuvm._reg.uvm_reg_model import (
+    uvm_access_e,
     uvm_coverage_model_e,
     uvm_door_e,
+    uvm_elem_kind_e,
+    uvm_status_e,
 )
 from pyuvm._s05_base_classes import uvm_object
 
 if TYPE_CHECKING:
-    from pyuvm._reg.uvm_mem_mam import (
-        alloc_mode_e,
-        locality_e,
-        uvm_mem_mam,
-        uvm_mem_mam_cfg,
-    )
+    from pyuvm._reg.uvm_mem_mam import uvm_mem_mam
     from pyuvm._reg.uvm_reg_backdoor import uvm_reg_backdoor
     from pyuvm._reg.uvm_reg_block import uvm_reg_block
-    from pyuvm._reg.uvm_reg_item import uvm_reg_item
     from pyuvm._reg.uvm_reg_map import uvm_reg_map, uvm_reg_map_info
     from pyuvm._reg.uvm_reg_model import (
         uvm_hdl_path_concat,
@@ -27,7 +27,6 @@ if TYPE_CHECKING:
         uvm_reg_addr_t,
         uvm_reg_cvr_t,
         uvm_reg_data_t,
-        uvm_status_e,
     )
     from pyuvm._reg.uvm_reg_sequence import uvm_reg_frontdoor
     from pyuvm._reg.uvm_vreg import uvm_vreg
@@ -67,7 +66,7 @@ class uvm_mem(uvm_object):
         self._vregs: list[uvm_vreg] = list()
         # self._hdl_paths_pool: uvm_object_string_pool = None
         self._mam: uvm_mem_mam = None
-        raise NotImplementedError
+        self._atomic = Lock()
 
     def configure(self, parent: uvm_reg_block, hdl_path: str = "") -> None:
         if not parent:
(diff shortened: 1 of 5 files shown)


pyuvm/_reg/uvm_mem.py
```diff
@@ -382,15 +457,109 @@ async def peek(
...
+        rw.set_status(uvm_status_e.UVM_IS_OK)
+        # TODO: pre_read callbacks
+        door = rw.get_door()
+        if door == uvm_door_e.UVM_BACKDOOR:
+            await self._do_read_backdoor(rw, map_info)
+        elif door == uvm_door_e.UVM_FRONTDOOR:
+            await self._do_read_frontdoor(rw, map_info)
+        # TODO: post_read callbacks
+        # TODO: report
+        self._read_in_progress = False
```
ktbarrett:
```suggestion
        try:
            self._read_in_progress = True
            rw.set_status(uvm_status_e.UVM_IS_OK)
            # TODO: pre_read callbacks
            door = rw.get_door()
            if door == uvm_door_e.UVM_BACKDOOR:
                await self._do_read_backdoor(rw, map_info)
            elif door == uvm_door_e.UVM_FRONTDOOR:
                await self._do_read_frontdoor(rw, map_info)
            # TODO: post_read callbacks
            # TODO: report
       finally:
            self._read_in_progress = False
```

ktbarrett (commented):
Looks like the tests could be expanded a bit. I wish we had coverage set up for this repo, but it looks like not all of the functionality introduced is tested.

Generally it looks good though.

raysalemi:
Hi Vishnu (@vs4848),

Will you be adding tests to this PR?

Ray

ktbarrett:
I think we can merge with my requested changes.

I'm putting together coverage collection for pyuvm and registering with [codecov.io](https://codecov.io) so we can track coverage in PRs going forward.

raysalemi:
But the changes aren't done yet, right?

pyuvm/_reg/uvm_mem.py
```diff
@@ -382,15 +457,111 @@ async def peek(
...
+            rw.set_status(uvm_status_e.UVM_IS_OK)
+            # TODO: pre_read callbacks
+            door = rw.get_door()
+            if door == uvm_door_e.UVM_BACKDOOR:
+                await self._do_read_backdoor(rw, map_info)
+            elif door == uvm_door_e.UVM_FRONTDOOR:
+                await self._do_read_frontdoor(rw, map_info)
+            # TODO: post_read callbacks
+            # TODO: report
+       finally:
```
ktbarrett:
```suggestion
        finally:
```

ktbarrett:
The coverage is too low. In fact there are some bugs I missed `atomic_lock` and `atomic_unlock` are not awaited. WE don't see the bug because the tests don't ever hit that code.

Until the coverage radically improves this will not be merged.

ktbarrett (changes requested):

raysalemi:
Was this PR superceded by 412?

Gemini says this about testing:

Conclusion
Fail. While the implementation logic accurately matches the PR's description and successfully lays out the foundation for uvm_mem, it completely fails the Test Sufficiency Matrix. A vast majority of the new conditional branches, edge cases, lock semantics, and actual frontdoor bus operations have zero corresponding test coverage. (Note: This directly contextualizes why PR #412 was later submitted to expand the patch coverage to 100%).

PR #396, opened 2026-07-14
[uvm_report, logging] Add opt-in SV-UVM-style reporting over Python logging

ysekhar:
# TL;DR

Adds an opt-in SV-UVM-style reporting layer while keeping Python logging as the
backend and preserving existing `self.logger.*` behavior by default. The new
path provides `self.uvm_report`, UVM severity and verbosity constants, report
IDs, UVM-style verbosity filtering, centralized counts, summaries, and basic
severity demotion. The Phase 5 hardening pass keeps global/default report
server verbosity explicit, makes report-server lifecycle state explicit, and
verifies direct Python logging is counted once through the shared `uvm` logger
path. It also keeps reused report-object loggers from carrying legacy per-object
stream handlers into the shared SV-UVM-style logger path. Formatting stays on
the Python logging / `PyuvmFormatter` path, and UVM verbosity remains separate
from Python logging levels.

# Summary

This PR adds the first SV-UVM-style reporting layer on top of pyuvm's existing
Python logging backend.

The default pyuvm logging behavior remains unchanged. Existing tests can keep
using `self.logger.info(...)`, `self.logger.warning(...)`, and
`self.logger.error(...)` without enabling the new reporting path.

When enabled, the new reporting path provides:

- UVM severity constants: `UVM_INFO`, `UVM_WARNING`, `UVM_ERROR`, `UVM_FATAL`
- UVM verbosity constants: `UVM_NONE`, `UVM_LOW`, `UVM_MEDIUM`, `UVM_HIGH`,
  `UVM_FULL`, `UVM_DEBUG`
- a lazy `uvm_object.uvm_report` reporter with
  `info()`, `warning()`, `error()`, and `fatal()` methods
- UVM-style verbosity filtering for info reports
- global/default report-server verbosity updates through `set_verbosity()`
- report IDs carried through logging records and formatter wrapping
- centralized `uvm_report_server` severity counts and summary output
- configurable fail policy and quit-count handling
- simple report catching through severity rewrite rules
- direct Python logging count integration when the report server is active
- top-level `pyuvm` exports for the new reporting symbols
- an opt-in shared reporting mode controlled by
  `PYUVM_ENABLE_SV_UVM_STYLE_REPORTING` or
  `set_sv_uvm_style_reporting_enabled(True)`
  
 # RFC Alignment

This implements the staged reporting direction discussed in
https://github.com/pyuvm/pyuvm/issues/374.

The RFC feedback called out two important design points:

[...]

Issue #374, opened 2026-06-17
RFC: Add UVM-Style Reporting Compatibility on Top of Python Logging


ysekhar:
## TL;DR

Add UVM-style reporting APIs, log patterns, report server counts, and basic report catching while preserving Python logging

## Summary

pyuvm currently gives each `uvm_report_object` a Python `logging.Logger` and expects users to report through normal Python logging calls such as `self.logger.info(...)`, `self.logger.warning(...)`, and `self.logger.error(...)`.

That is useful and Pythonic, but it leaves a compatibility gap for users moving UVM-style verification code or methodology into pyuvm. Common UVM reporting features are currently absent, including:

- `uvm_report.info`, `uvm_report.warning`, `uvm_report.error`, and `uvm_report.fatal`
- UVM verbosity levels such as `UVM_LOW`, `UVM_MEDIUM`, and `UVM_HIGH`
- report IDs
- UVM-like log line patterns for external regression/orchestration tools
- severity counts
- quit count
- report summary
- report catching / severity demotion
- centralized report server behavior

This RFC proposes adding a staged UVM-style reporting layer to pyuvm while keeping Python logging as the backend and preserving all existing logger-based usage. One important goal is to make pyuvm logs easier for existing verification orchestration infrastructure to consume, because many regression systems already look for UVM-style severity patterns and report summaries.

## Motivation

The current implementation is intentionally minimal and Python-native. That should remain true. However, there are practical reasons to add a UVM-compatible reporting layer:

1. Many UVM users expect report IDs, verbosity filtering, severity counts, and report summaries.
2. Existing UVM-style code ports less cleanly when every report must be rewritten as a raw Python logging call.
3. Verification regressions often need deterministic pass/fail behavior based on `UVM_ERROR` and `UVM_FATAL` counts.
4. Message demotion and severity rewrite are important in larger DV environments.
5. Python logging can remain the transport/backend while pyuvm provides UVM-shaped reporting semantics.
6. Existing verification orchestration infrastructure often parses UVM reporting patterns in logs. Emitting familiar severity tokens and summaries would lower adoption friction for teams that already have UVM-oriented CI, dashboards, log scanners, or regression managers.

The goal is not to emulate every SystemVerilog reporting mechanism immediately. The goal is to provide the useful compatibility surface in a Python-friendly way.

## Current Behavior

`uvm_report_object` currently:

- inherits from `uvm_object`
- creates a Python logger under the `uvm` root logger
- attaches a stream handler
[...]

PR #398, opened 2026-07-15
[examples, TinyALU] Add TinyALU UVM-style reporting example

ysekhar:
# TL;DR

Adds a TinyALU example for pyuvm's opt-in SV-UVM-style reporting path. The
example demonstrates report IDs, source file/line output, UVM hierarchy,
verbosity, report summaries, `TEST_STATUS`, max quit count handling, fatal
reports, and severity demotion using real TinyALU scoreboard traffic.

The pyuvm reporting changes in this PR are the support needed to make that
example work cleanly from a cocotb regression: centralized report-server setup
from `uvm_test`, runtime options from plusargs/env vars, graceful max-quit
termination, customizable test-status labels, and test-specific message
demotions.

# Summary

This PR adds `examples/TinyALU_uvm_reporting`, a TinyALU regression focused on
SV-UVM-style reporting behavior.

The existing `examples/TinyALU` remains the direct Python logging walkthrough.
The new example uses the same TinyALU RTL/BFM flow, but reports through
`self.uvm_report`.

The example demonstrates:

- passing ALU traffic with UVM report summaries
- source file and line output on UVM-style report records
- report IDs and UVM hierarchy in log output
- sequence reporting through `self.uvm_report`
- scoreboard mismatch reports
- fatal reports from the same mismatch path
- max quit count handling at `0`, `1`, `2`, and `25`
- `+max_quit_count=0` as unlimited, reported as `UNLIMITED`
- graceful summary and final status output after max quit count is reached
- default `TEST_STATUS` output with a customizable label
- severity demotion from fatal/error reports to lower severities
- `[SEVCHG]` diagnostics when a report is demoted

# RFC Alignment

This builds on the SV-UVM-style reporting direction from #396 and the RFC #374 

The main goal here is not to replace pyuvm's Python logging model. It is to give
users a concrete example showing how the new reporting layer behaves in a real
verification-style testbench.

The TinyALU example exercises the pieces that teams commonly need in regression logs:
- report IDs
- source locations
- UVM hierarchy
- per-test pass/fail status
- max quit count behavior
- severity demotion hooks
- report summaries at test end

# Deliberate Differences from SystemVerilog UVM

This remains a Python-native pyuvm example, not a complete SystemVerilog UVM
reporting clone.

Important differences:

- Python logging remains the backend.
- Direct Python logger records are still ordinary logging records.
- UVM-style report behavior applies to `self.uvm_report.*` calls.
- Fatal reports use Python exception behavior.
- Max quit count exits gracefully through pyuvm/cocotb test flow.
- Severity demotion is intentionally focused on report ID/message matching.

# Compatibility

The new TinyALU reporting example enables SV-UVM-style reporting for itself.
Existing examples and tests using `self.logger.*` continue to work as before.

Runtime reporting options can be supplied through cocotb plusargs or matching
environment variables.

For example:

```sh
make SIM=verilator TOPLEVEL_LANG=verilog \
    COCOTB_PLUSARGS="+UVM_VERBOSITY=DEBUG +max_quit_count=25 +test_status_label=MY_TEST_STATUS +print_char_len=80" \
    sim checkclean
```

The example also adds a `maxquit-demo` make target to show the same error test
at multiple quit-count settings:

```sh
make SIM=verilator TOPLEVEL_LANG=verilog maxquit-demo
```
# Implementation Notes

- `examples/TinyALU_uvm_reporting` adds the new example, Makefile, README, and
  cocotb/pyuvm testbench.
- `uvm_test` now creates centralized reporting support when SV-UVM-style
  reporting is enabled.
- Runtime reporting options are resolved from plusargs first, then environment
  variables.
- `uvm_report_policy` defaults the final status label to `TEST_STATUS`, while
  allowing teams to customize it.
- Max quit count now supports unlimited mode with `0`.
- Max quit count emits the triggering report, emits the `[MAXQUIT]` message, and
  still prints the report summary/final status.
- `add_message_demotes(self, catcher)` lets a test register report demotions.
- Severity demotions emit a `[SEVCHG]` diagnostic at the report call site.
- UVM-style report records include useful source file/line information.

# Tests Added

The TinyALU reporting regression includes:

- `AluTest`
- `ParallelTest`
- `FibonacciTest`
- `AluTestErrors`
- `FatalReportTest`
- `FatalDowngradeToErrorTest`
- `FatalDowngradeToInfoTest`
- `ErrorDowngradeToInfoTest`

The pyuvm unit tests cover:

- report-server setup from `uvm_test`
- plusarg/env runtime option parsing
- plusarg override behavior
- default and customized `TEST_STATUS` labels
- max quit count at `0`, `1`, `2`, and larger values
- unlimited quit-count summary formatting
- graceful max-quit summary/final-status behavior
- severity demotion diagnostics
- fatal-to-error, fatal-to-info, and error-to-info demotions
- the `add_message_demotes()` hook
```

### DIFF
--- a/examples/TinyALU_uvm_reporting/testbench.py
+++ b/examples/TinyALU_uvm_reporting/testbench.py
@@ -1,0 +1,391 @@
+import random
+
+# All testbenches use tinyalu_utils, so store it in a central
+# place and add its path to the sys path so we can import it.
+import sys
+from pathlib import Path
+
+import cocotb
+from cocotb.clock import Clock
+from cocotb.triggers import Combine
+
+import pyuvm
+from pyuvm import *
+
+set_sv_uvm_style_reporting_enabled(True)
+
+sys.path.append(str(Path("../TinyALU").resolve()))
+from tinyalu_utils import Ops, TinyAluBfm, alu_prediction  # noqa: E402
+
+
+# Sequence classes
+
+
+class AluSeqItem(uvm_sequence_item):
+    def __init__(self, name, aa, bb, op):
+        super().__init__(name)
+        self.A = aa
+        self.B = bb
+        self.op = Ops(op)
+
+    def randomize_operands(self):
+        self.A = random.randint(0, 255)
+        self.B = random.randint(0, 255)
+
+    def randomize(self):
+        self.randomize_operands()
+        self.op = random.choice(list(Ops))
+
+    def __eq__(self, other):
+        same = self.A == other.A and self.B == other.B and self.op == other.op
+        return same
+
+    __hash__: None  # type: ignore
+
+    def __str__(self):
+        return f"{self.get_name()} : A: 0x{self.A:02x} \
+        OP: {self.op.name} ({self.op.value}) B: 0x{self.B:02x}"
+
+
+class RandomSeq(uvm_sequence):
+    async def body(self):
+        for op in list(Ops):
+            cmd_tr = AluSeqItem("cmd_tr", None, None, op)
+            await self.start_item(cmd_tr)
+            cmd_tr.randomize_operands()
+            await self.finish_item(cmd_tr)
+
+
+class MaxSeq(uvm_sequence):
+    async def body(self):
+        for op in list(Ops):
+            cmd_tr = AluSeqItem("cmd_tr", 0xFF, 0xFF, op)
+            await self.start_item(cmd_tr)
+            await self.finish_item(cmd_tr)
+
+
+class TestAllSeq(uvm_sequence):
+    async def body(self):
+        seqr = ConfigDB().get(None, "", "SEQR")
+        random = RandomSeq("random")
+        max = MaxSeq("max")
+        await random.start(seqr)
+        await max.start(seqr)
+
+
+class TestAllForkSeq(uvm_sequence):
+    async def body(self):
+        seqr = ConfigDB().get(None, "", "SEQR")
+        random = RandomSeq("random")
+        max = MaxSeq("max")
+        random_task = cocotb.start_soon(random.start(seqr))
+        max_task = cocotb.start_soon(max.start(seqr))
+        await Combine(random_task, max_task)
+
+
+# Sequence library example
+
+
+class OpSeq(uvm_sequence):
+    def __init__(self, name, aa, bb, op):
+        super().__init__(name)
+        self.aa = aa
+        self.bb = bb
+        self.op = Ops(op)
+
+    async def body(self):
+        seq_item = AluSeqItem("seq_item", self.aa, self.bb, self.op)
+        await self.start_item(seq_item)
+        await self.finish_item(seq_item)
+        self.result = seq_item.result
+
+
+async def do_add(seqr, aa, bb):
+    seq = OpSeq("seq", aa, bb, Ops.ADD)
+    await seq.start(seqr)
+    return seq.result
+
+
+async def do_and(seqr, aa, bb):
+    seq = OpSeq("seq", aa, bb, Ops.AND)
+    await seq.start(seqr)
+    return seq.result
+
+
+async def do_xor(seqr, aa, bb):
+    seq = OpSeq("seq", aa, bb, Ops.XOR)
+    await seq.start(seqr)
+    return seq.result
+
+
+async def do_mul(seqr, aa, bb):
+    seq = OpSeq("seq", aa, bb, Ops.MUL)
+    await seq.start(seqr)
+    return seq.result
+
+
+class FibonacciSeq(uvm_sequence):
+    def __init__(self, name):
+        super().__init__(name)
+        self.seqr = ConfigDB().get(None, "", "SEQR")
+
+    async def body(self):
+        prev_num = 0
+        cur_num = 1
+        fib_list = [prev_num, cur_num]
+        for _ in range(7):
+            sum = await do_add(self.seqr, prev_num, cur_num)
+            fib_list.append(sum)
+            prev_num = cur_num
+            cur_num = sum
+        self.uvm_report.info(
+            self.get_name(), "Fibonacci Sequence: " + str(fib_list), UVM_LOW
+        )
+
+
+class Driver(uvm_driver):
+    def build_phase(self):
+        self.ap = uvm_analysis_port("ap", self)
+
+    def start_of_simulation_phase(self):
+        self.bfm = TinyAluBfm()
+
+    async def launch_tb(self):
+        await self.bfm.reset()
+        self.bfm.start_bfm()
+
+    async def run_phase(self):
+        await self.launch_tb()
+        while True:
+            cmd = await self.seq_item_port.get_next_item()
+            await self.bfm.send_op(cmd.A, cmd.B, cmd.op)
+            result = await self.bfm.get_result()
+            self.ap.write(result)
+            cmd.result = result
+            self.seq_item_port.item_done()
+
+
+class Coverage(uvm_subscriber):
+    def end_of_elaboration_phase(self):
+        self.cvg = set()
+
+    def write(self, cmd):
+        (_, _, op) = cmd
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: examples/TinyALU_uvm_reporting/testbench.py [hunk 1]
```diff
@@ -0,0 +1,386 @@
+import random
```
ktbarrett: I'm not sure it's worthwhile to copy paste the entire thing just to enable the logging, why not set the logging in the other test?

pyuvm/uvm_reporting/uvm_report_server.py
```diff
@@ -345,41 +329,48 @@ def failure_message(self, context: str = "") -> str | None:
...
+        self._log_uvm_info(logger, "** Report counts by severity", full_name)
+        self._log_uvm_info(logger, f"UVM_INFO    : {stats.info_count:4d}", full_name)
+        self._log_uvm_info(logger, f"UVM_WARNING : {stats.warning_count:4d}", full_name)
+        self._log_uvm_info(logger, f"UVM_ERROR   : {stats.error_count:4d}", full_name)
+        self._log_uvm_info(logger, f"UVM_FATAL   : {stats.fatal_count:4d}", full_name)
+        max_quit_count = self._max_quit_count_text()
+        self._log_uvm_info(
+            logger,
+            f"Quit count : {stats.error_quit_count:5d} of {max_quit_count} (UVM_ERROR)",
+            full_name,
```
ktbarrett, line 353: Are multiline logs not allowed?

ktbarrett (commented):
Python-wise it's pretty solid. Whether this is more UVM like or not @raysalemi.

ktbarrett (approved):

examples/TinyALU_uvm_reporting/testbench.py
raysalemi: Not sure what you mean here.

pyuvm/uvm_reporting/uvm_report_server.py
raysalemi: UVM messages are usually one line.  Though I imagine \n could be inserted.

examples/TinyALU_uvm_reporting/testbench.py
ktbarrett: This seems like duplicate of the TinyALU example with just a couple lines added.

raysalemi: Well, that’s true.  Perhaps better to extend the component and override it in the factory.

raysalemi: @ysekhar , thoughts?

ysekhar: @raysalemi let me experiment. The reason I built a separate classes was because I will need to modify all the logger statements in the original example and wrap them so that the inherited classes can redirect the wrapper function to the uvm-reporter. 

The way I did it seemed simpler. Can check in a commit with that implementation and see if it is acceptable. If not we can drop it before merge to master

raysalemi:
I always look twice at copied code.  It is hard to maintain.

That said, it is simpler.

ysekhar:
@raysalemi @ktbarrett, I have pushed a commit that works on inheritance and factory overrides for the TinyALU example. 

The original TinyALU example was updated only to make its existing behavior reusable through normal pyuvm factory/subclassing patterns.

The behavioral flow is unchanged. The same sequences run, the same driver/monitor/coverage/scoreboard topology is used, and the original example still reports through `self.logger`.

The changes are:

- Small reporting hook methods were added in `FibonacciSeq`, `Coverage`, `Scoreboard`, and `Monitor`. These keep the original logger behavior in the base example, but allow the UVM reporting example to override only the reporting calls.
- The repeated scoreboard message formatting was moved into helper methods so the reporting subclass can reuse the same PASS/FAILED text.
- `CommandMonitor`, `Coverage`, `Scoreboard`, and `AluEnv` are now created through `create()` so the reporting example can install factory overrides instead of copying the whole testbench.
- `Monitor` still does the same command monitoring; `CommandMonitor` just packages the existing `get_cmd` construction into a factory-creatable component.

This lets `examples/TinyALU_uvm_reporting/testbench.py` shrink to a thin adapter over the original TinyALU example rather than a near-copy of it.

pyuvm/uvm_reporting/uvm_report_server.py
```diff
@@ -442,6 +432,111 @@ def _apply_catcher(self, msg: str, severity: str, report_id: str = "") -> str:
...
+            return registered
+
+        name = str(getattr(logger, "name", "") or "")
+        if name == "uvm":
+            return "uvm"
+        if not name.startswith("uvm."):
+            return ""
+        full_name = name[4:]
+        if full_name and full_name[-1].isdigit():
+            full_name = full_name.rstrip("0123456789")
```
ktbarrett: Is there a reason to strip the numbers?

pyuvm/_s06_reporting_classes.py
```diff
@@ -110,8 +110,7 @@ def __init__(self, name):
             configure_uvm_root_logger()
             self.logger.propagate = True
             for handler in list(self.logger.handlers):
-                if getattr(handler, "_pyuvm_object_default_handler", False):
-                    self.logger.removeHandler(handler)
```
ktbarrett, line 114: I'm not sure this is necessary, or we need some other way to disambiguate pyuvm handlers from all others. Maybe a base class for pyuvm's handlers?

examples/TinyALU_uvm_reporting/testbench.py
```diff
@@ -0,0 +1,190 @@
...
+    tinyalu_base = sys.modules[_BASE_MODULE_NAME]
+else:
+    _base_spec = importlib.util.spec_from_file_location(
+        _BASE_MODULE_NAME, _BASE_MODULE_PATH
+    )
+    if _base_spec is None or _base_spec.loader is None:
+        raise ImportError(f"Could not load TinyALU testbench from {_BASE_MODULE_PATH}")
+    tinyalu_base = importlib.util.module_from_spec(_base_spec)
+    sys.modules[_BASE_MODULE_NAME] = tinyalu_base
+    _base_spec.loader.exec_module(tinyalu_base)
```
ktbarrett, line 27: Yeah this isn't fun, but we will clean this up in a follow on (not your responsibility) by reorganizing the test and examples dirs to be Python namespace packages added to the PYTHONPATH so we can just `from examples.TinyALU.testbench import ...`.

pyuvm/_s06_reporting_classes.py
ktbarrett: We can punt on this for now. #405

examples/TinyALU_uvm_reporting/testbench.py
ktbarrett: Punted: #406.

pyuvm/uvm_reporting/uvm_report_server.py
ysekhar: This is now removed. I should have cleaned up some debug code before the last push.

pyuvm/uvm_reporting/uvm_report_server.py
ysekhar: Added support for multiline message when using uvm_reporting

ktbarrett (approved):

ysekhar:
@raysalemi @ktbarrett let me know if there is anything else gating this PR being merged.

raysalemi:
Sorry. I will do it now

ktbarrett:
Yeah my bad, I have a lot going on and needed a reminder.

ysekhar:
Thanks you both for patiently shepherding this PR. 
Look forward to more conversations as we work to make pyuvm the default library.

raysalemi:
Isn't this merged?

Issue #403, opened 2026-07-17
Bug Report


ktbarrett:
Attached below is a Code Review performed by Claude 4.8 on the existing code in the library as of pyuvm/pyuvm@d133a556bfee30f1961ceeb24c9d8f16347a1865.

<details><summary>Code Review</summary>

# pyuvm Code Review

Thorough review of the pyuvm codebase (a Python implementation of IEEE 1800.2-2020
UVM on top of cocotb) against the spec text (`1800.2-2020`). Reviewed end-to-end:
all of `pyuvm/` including the register layer (`pyuvm/_reg/`) and reporting
(`pyuvm/uvm_reporting/`).

**Scope notes / ground rules honored during review**
- Intentional divergences to fit Python semantics (notably sequences in
  `_s14_15_python_sequences.py`, which are a fresh design) are *not* flagged as
  spec-fidelity bugs; only genuine internal-correctness problems in that area are
  reported.
- Missing type annotations are intentional and are **not** reported.
- Many register-layer methods are deliberate `NotImplementedError` scaffolding;
  those stubs are not flagged. Bugs are only reported in code clearly meant to run.

## How to read this

Findings are grouped by **category** and ordered by **severity** within each group.
Each entry cites `file:line`. A severity-ordered index is at the top; per-module
raw notes live in `/tmp/pyuvm_review/*.md`.

Severity meaning:
- **Critical** — silently wrong results or hangs on a common/normal path.
- **High** — broken behavior on a real (if less common) path, or a crash where the
  code is meant to work.
- **Medium** — wrong behavior in specific configurations, or latent crash behind a
  currently-unused path.
- **Low** — cosmetic, diagnostic, maintainability, or dormant/dead-code issues.

---

### Critical (4)
1. `uvm_reg_field._predict`: **W1C/W1CRS write-prediction clears ALL bits** — `_reg/uvm_reg_field.py:511`
2. `uvm_callbacks.add`: **duplicate add wipes the entire callback queue** — `_s10_synchronization_classes.py:62`
3. `uvm_tlm_transport_channel.transport`: **request `put` not awaited → transport channel hangs** — `_s12_uvm_tlm_interfaces.py:1018`
4. `uvm_root.run_test`: **clears its own singleton then keeps using the stale `self`** — `_s13_uvm_component.py:487`

### High (14)
5. Five common phases have wrong top-down/bottom-up direction — `_s09_phasing.py:117,121,129,133,137`
[...]

PR #404, opened 2026-07-17
Various Fixes

ktbarrett:
xref #403.

The objection mechanism uses counts instead of a flag, I had claude cook up a fix for this. Also started tackling some of the issues in the Bug Report.


raysalemi:
But the spec calls for a count, right?

ktbarrett:
@raysalemi Yes exactly, it was a flag before, and now it's a count. Really there should be a timeout mechanism as well, but going piecemeal.

pyuvm/_s05_base_classes.py
```diff
@@ -637,22 +643,31 @@ def get_event_pool(self):
     def get_accept_time(self) -> int:
         """
         :return: Accept time of transaction
+        :raises UVMError: if ``accept_tr()`` was never called
 
         """
+        if self._accept_time is None:
+            raise UVMError(f"accept_tr() was never called on {self.get_name()}")
```
ktbarrett, line 650: These I'm not sure if this is the correct answer. In SV/UVM this returns a time object of value -1, which is invalid. I'm not sure if `int | None` or raising an exception is more appropriate here instead.

raysalemi:
Raising an exception is more appropriate in this case.  There is no going
forward.

Ray

PR #402, opened 2026-07-16
[pyral, core] Add foundational RAL register and frontdoor access APIs

ysekhar:
# TL;DR

Implements the first four phases of the practical pyuvm RAL RFC: baseline register-model behavior tests, core `uvm_reg` APIs, field access helpers, and a working frontdoor register read/write path through `uvm_reg_map` and `uvm_reg_adapter`.

This makes simple pyuvm register models usable for construction, field packing, desired/mirrored/reset state, register update/mirror flows, field read/write delegation, and mock-adapter frontdoor access without hitting the previous core stubs.

# Summary

This PR adds the foundational RAL behavior needed before memory, full backdoor dispatch, register sequences/frontdoors, and predictor work can land.

It covers:

- register/block/map construction and address lookup behavior
- register field ordering, overlap reporting, packing, and prediction
- `uvm_reg` helpers for field lookup, offsets, masks, reset state, update, mirror, compare, peek/poke item flow, and frontdoor/backdoor setter/getter APIs
- `uvm_reg_field` helpers for masks, register insertion/extraction, direct field read/write, peek/poke, mirror, update values, and access-policy prediction
- register-map frontdoor read/write dispatch through a sequencer and adapter
- bus op propagation for kind, address, data, bit width, byte enables, and status
- auto-prediction only after successful frontdoor accesses
- useful fatal errors for missing adapter/sequencer setup
- canonical `provides_responses` adapter spelling, with `provides_response` kept as a compatibility property

# RFC Alignment

This follows the staged RAL direction from the RFC #399 

Covered in this PR (single commit per phase):

- Phase 1: lock down existing register model behavior
- Phase 2: complete practical core `uvm_reg` APIs
- Phase 3: complete practical `uvm_reg_field` access and packing helpers
- Phase 4: harden frontdoor register access through map and adapter

The intent is not to complete the entire IEEE register layer in one PR. This change establishes the tested register and adapter foundation that later RAL work can build on.

# Deliberate Scope Boundaries

This PR does not implement the remaining RAL phases:

- full base-class backdoor dispatch remains follow-on work
- `uvm_mem` construction, mapping, and memory access remain follow-on work (Should leverage what is done in #393)
- `uvm_reg_sequence` / `uvm_reg_frontdoor` usability remains follow-on work
- `uvm_reg_predictor` remains follow-on work
- HDL path behavior remains storage/introspection-oriented for now

[...]

Issue #399, opened 2026-07-15
RFC: Complete a Practical pyuvm Register Abstraction Layer Core


ysekhar:
## Summary

pyuvm already includes a substantial `_reg` package with IEEE 1800.2-shaped
classes and canonical pyuvm enum names. However, several core register-layer
operations are still placeholders or only partially connected. This makes the
current RAL useful for structural model construction and some prediction
experiments, but not yet sufficient for ordinary register-model access through
frontdoor or backdoor paths.

This RFC proposes completing a practical, tested RAL core in stages:

1. Preserve and test the existing register/block/map construction behavior.
2. Complete core `uvm_reg` and `uvm_reg_field` operations.
3. Harden register frontdoor access through `uvm_reg_map`, `uvm_reg_adapter`,
   and `uvm_reg_bus_op`.
4. Add a minimal Python backdoor path for registers.
5. Bring `uvm_mem` and memory map support to parity with simple registers.
6. Make `uvm_reg_sequence` and `uvm_reg_frontdoor` usable by derived
   sequences.
7. Implement `uvm_reg_predictor` after the direct access path is stable.
8. Document the supported subset and explicitly mark longer-tail gaps.

The goal is not to implement every IEEE register-layer feature in one change.
The goal is to make a small but coherent RAL subset work end-to-end, with tests
and clear public behavior.

## Motivation

Register models are a common part of UVM-style verification environments.
pyuvm already exposes many familiar RAL class names, including:

- `uvm_reg_block`
- `uvm_reg_map`
- `uvm_reg`
- `uvm_reg_field`
- `uvm_mem`
- `uvm_reg_adapter`
- `uvm_reg_bus_op`
- `uvm_reg_sequence`
- `uvm_reg_frontdoor`
- `uvm_reg_backdoor`
- `uvm_reg_predictor`

That public surface suggests users should be able to build a small register
model and perform reads and writes through normal pyuvm sequences and adapters.
Today, that workflow still runs into missing methods in important places.

Completing the practical RAL core would:

1. Make pyuvm more useful for realistic register-oriented testbenches.
2. Reduce the need for local project-specific RAL shims.
3. Give users a tested path for register read/write, mirror/update, peek/poke,
   and memory access.
4. Provide a foundation for future IEEE register-layer coverage without forcing
   every feature into the first implementation.
5. Clarify which RAL APIs pyuvm supports and which remain intentionally
   incomplete.

## Proposed Direction

[...]

PR #412, opened 2026-07-27
[pyral, core] Implement RAL memory mapping and frontdoor access

ysekhar:
# TL;DR

Adds constructible RAL memories, block and map integration, exact memory address lookup, and single-element frontdoor reads and writes through the existing register-map adapter and sequencer path.

Mapped memories use compact, stride-aware address descriptors, allowing large memories to be represented without creating one Python lookup entry per memory location.

This PR supersedes #393 with exact address semantics, shared register/memory bus transport, and 100% patch coverage.

# Summary

This PR builds on the register and frontdoor foundation merged in #402 the practical memory behavior needed for pyuvm register models.

It covers:
- constructible `uvm_mem` objects with validated size, element width, and access policy
- memory registration and lookup through `uvm_reg_block`
- memory mapping through `uvm_reg_map.add_mem`
- map offset, rights, unmapped state, element stride, address metadata, and initialization state
- `uvm_mem.get_offset`, `get_address`, and `get_addresses`
- byte-addressed, word-addressed, and hierarchical memory address translation
- exact address lookup through `uvm_reg_map.get_mem_by_offset`
- exact register/memory and memory/memory overlap diagnostics
- dynamic memory remapping and unmapping after the model is locked
- single-element memory reads and writes through the configured adapter and sequencer
- map-level `RW`, `RO`, and `WO` access restrictions
- bounds and unmapped-access validation before adapter activity
- propagation of bus address, data, bit width, byte enable, and response status
- atomic serialization using `cocotb.triggers.Lock`
- SV-UVM-style warning and error reporting when enabled
- recoverable UVM address-query behavior for invalid or unavailable addresses

# RFC Alignment

This continues the staged RAL direction introduced in RFC #399 and builds on the foundation merged in #402.

The implementation is split into three commits:
1. Implement RAL memory support
2. Make RAL memory mapping scalable
3. Refine memory address introspection

The intent is not to complete every UVM memory feature in one PR. This change establishes a tested and scalable memory foundation for later frontdoor, backdoor, sequence, predictor, and virtual-register work.

# PR #393

This PR supersedes #393.

PR #393 established the initial direction for `uvm_mem` construction, map plumbing, and frontdoor access. This implementation carries that work forward while addressing the issues identified during review:
- expands patch coverage from 27% to 100%
- uses awaited cocotb locks for atomic memory operations
- validates bounds, map rights, unmapped state, adapter setup, and sequencer setup before bus activity
- shares bus-operation construction and response handling with register access
- preserves holes in strided address maps
- detects register/memory overlaps as well as memory/memory overlaps
- avoids lookup storage proportional to memory depth

The burst and custom-frontdoor implementations from #393 are not included because those features require their own transaction, callback, and failure semantics.

# Scalable Address Mapping

A memory location may occupy one or more physical bus addresses, and consecutive locations may be separated by an address stride.

A simple minimum/maximum range is insufficient because it treats holes between strided locations as occupied. Expanding every location into a dictionary is exact, but consumes memory proportional to the modeled memory depth.

This PR instead stores a compact descriptor for each mapped memory:
- the physical address of each byte lane for element zero
- the address stride of each lane between elements
- the number of memory elements
- minimum and maximum bounds for fast rejection

Lookup membership is calculated mathematically. Memory/memory overlap is detected by finding intersections between finite arithmetic progressions using GCD and modular arithmetic.

The resulting metadata is proportional to the number and width of mapped memories rather than their depth. A regression locks and queries a 10-million-element memory without creating 10 million lookup entries.

# UVM Behavior

The implementation follows the relevant UVM memory and map behavior within the supported scope:
- intrinsic memory access is `RW` or `RO`
- per-map rights may be `RW`, `RO`, or `WO`
- memories may appear in multiple address maps
- logical memory offsets are translated into external physical addresses
- `get_mem_by_offset` is available after model locking
- memories do not maintain desired or mirrored state
- failed address-introspection queries are recoverable and use the UVM `-1`
  sentinel rather than terminating the simulation
- invalid operations return `UVM_NOT_OK` without invoking the adapter

Memory elements wider than the map bus are currently rejected with a useful diagnostic. Generic multi-beat memory transport is deferred rather than being partially implemented.

# Deliberate Scope Boundaries

This PR does not implement:

- burst reads and writes
- user-defined memory frontdoor sequences
- `peek` and `poke`
- base memory backdoor dispatch
- inherited block backdoors
- functional HDL-path access
- memory allocation manager and regions
- virtual registers and virtual fields
- memory callbacks and functional coverage
- sparse desired or mirrored memory state
- generic multi-beat memory elements

These remain follow-on RAL work.

# Compatibility

The implementation preserves pyuvm's existing enum and transaction conventions, including:
- `uvm_status_e`
- `uvm_door_e`
- `uvm_access_e`
- `uvm_elem_kind_e`
- `uvm_reg_item`
- `uvm_reg_bus_op`

Existing register frontdoor behavior continues to use the same shared adapter and sequencer machinery.

Standard `RegModel` logging remains the default. When SV-UVM-style reporting is enabled, new memory diagnostics are routed through `uvm_report` with stable report IDs.

# Implementation Notes

- `pyuvm/_reg/uvm_mem.py`
  - validates and constructs memories
  - registers memories with their parent blocks
  - implements address introspection
  - creates memory access items
  - validates bounds, rights, mappings, and supported width
  - performs single-element frontdoor reads and writes
  - routes diagnostics through the configured reporting system

- `pyuvm/_reg/uvm_reg_block.py`
  - adds private memory registration analogous to register registration

- `pyuvm/_reg/uvm_reg_map.py`
  - implements memory map configuration and lookup
  - adds compact stride-aware memory address descriptors
  - detects exact register/memory and memory/memory overlaps
  - rebuilds memory metadata after remapping
  - generalizes bus-operation construction for registers and memories
  - centralizes request/response handling through the adapter and sequencer

- `pyuvm/_reg/uvm_reg_model.py`
  - adds immutable memory address-range metadata

# Tests Added

New focused RAL memory coverage includes:

- `tests/pytests/reg/test_uvm_mem_construction.py`
- `tests/pytests/reg/test_uvm_mem_mapping.py`
- `tests/pytests/reg/test_uvm_mem_frontdoor.py`

The tests cover:

- valid and invalid memory construction
- block registration and hierarchical lookup
- standard and SV-UVM-style diagnostics
- map membership, rights, remapping, and unmapping
- byte-addressed and word-addressed maps
- hierarchical map translation
- exact stride-hole lookup
- register/memory and memory/memory overlaps
- single-element and very large memories
- bounds and access-policy rejection
- adapter request and response handling
- byte enables, data width, address, value, and status propagation
- missing adapter and sequencer diagnostics
- atomic memory access behavior

### DIFF
--- a/pyuvm/_reg/uvm_mem.py
+++ b/pyuvm/_reg/uvm_mem.py
@@ -40,6 +40,13 @@ __all__ = ["uvm_mem"]
 logger = logging.getLogger("RegModel")
 
 
+def _report_warning(obj: uvm_object, report_id: str, msg: str) -> None:
+    if get_sv_uvm_style_reporting_enabled():
+        obj.uvm_report.warning(report_id, msg)
+    else:
+        logger.warning(msg)
+
+
 def _report_error(obj: uvm_object, report_id: str, msg: str) -> None:
     if get_sv_uvm_style_reporting_enabled():
         obj.uvm_report.error(report_id, msg)
@@ -179,15 +186,19 @@ class uvm_mem(uvm_object):
                 if parent_map == map:
                     return local_map
                 parent_map = parent_map.get_parent_map()
-        logger.warning(
+        _report_warning(
+            self,
+            "MEM_MAP_LOOKUP",
             f"Memory '{self.get_full_name()}' is not contained "
-            f"within map '{map.get_full_name()}'"
+            f"within map '{map.get_full_name()}'",
         )
 
     def get_default_map(self) -> uvm_reg_map | None:
         if not self._maps:
-            logger.warning(
-                f"Memory '{self.get_full_name()}' is not registered with any map"
+            _report_warning(
+                self,
+                "MEM_NO_MAP",
+                f"Memory '{self.get_full_name()}' is not registered with any map",
             )
             return
         if len(self._maps) == 1:
@@ -268,17 +279,21 @@ class uvm_mem(uvm_object):
         for vreg in self.get_virtual_registers():
             if vreg.get_name() == name:
                 return vreg
-        logger.warning(
+        _report_warning(
+            self,
+            "MEM_VREG_LOOKUP",
             f"Unable to find virtual register '{name}' in memory "
-            f"'{self.get_full_name()}'"
+            f"'{self.get_full_name()}'",
         )
 
     def get_vfield_by_name(self, name: str) -> uvm_vreg_field | None:
         for field in self.get_virtual_fields():
             if field.get_name() == name:
                 return field
-        logger.warning(
-            f"Unable to find virtual field '{name}' in memory '{self.get_full_name()}'"
+        _report_warning(
+            self,
+            "MEM_VFIELD_LOOKUP",
+            f"Unable to find virtual field '{name}' in memory '{self.get_full_name()}'",
         )
 
     def get_vreg_by_offset(
@@ -290,9 +305,11 @@ class uvm_mem(uvm_object):
         self, offset: uvm_reg_addr_t = 0, map: uvm_reg_map = None
     ) -> uvm_reg_addr_t:
         if offset < 0 or offset >= self._size:
-            logger.warning(
+            _report_warning(
+                self,
+                "MEM_OFFSET",
                 f"Offset 0x{offset:X} lies outside of memory "
-                f"'{self.get_name()}' which has size 0x{self._size:X}"
+                f"'{self.get_name()}' which has size 0x{self._size:X}",
             )
             return -1
         local_map = self.get_local_map(map)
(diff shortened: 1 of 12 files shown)


pyuvm/_reg/uvm_mem.py
```diff
@@ -269,7 +294,19 @@ def get_vreg_by_offset(
     def get_offset(
         self, offset: uvm_reg_addr_t = 0, map: uvm_reg_map = None
     ) -> uvm_reg_addr_t:
-        raise NotImplementedError
+        if offset < 0 or offset >= self._size:
+            logger.warning(
+                f"Offset 0x{offset:X} lies outside of memory "
+                f"'{self.get_name()}' which has size 0x{self._size:X}"
+            )
+            return -1
```
ktbarrett: Here is the `return -1` thing here again. This should either raise and exception or return `None` and the type annotation should be `uvm_reg_addr_t | None`.

pyuvm/_reg/uvm_mem.py
```diff
@@ -307,33 +324,47 @@ def get_address(
         self, offset: uvm_reg_addr_t = 0, map: uvm_reg_map = None
     ) -> uvm_reg_addr_t:
         _, addresses = self.get_addresses(offset, map)
-        return addresses[0]
+        # UVM address-introspection APIs use -1 as the recoverable failure
+        # sentinel when an offset is invalid or no mapped address is available.
+        # This is not an exception because a failed address query does not
+        # corrupt the model or prevent simulation from continuing; exceptions
+        # are reserved for fatal configuration and model-integrity failures.
+        return addresses[0] if addresses else -1
```
ktbarrett: `-1` is not a `uvm_reg_addr_t`. Python does not auto construct values into the annotated type. If this is non-fatal, this function should be returning `None` and the type should be `uvm_reg_addr_t | None` as that's more idiomatic Python and more in line with the intent.

pyuvm/_reg/uvm_reg_block.py
```diff
@@ -39,6 +40,20 @@
...
+        obj.uvm_report.warning(report_id, msg)
+    else:
+        logger.warning(msg)
+
+
+def _report_error(obj: uvm_object, report_id: str, msg: str) -> None:
+    if get_sv_uvm_style_reporting_enabled():
+        obj.uvm_report.error(report_id, msg)
+    else:
+        logger.error(msg)
```
ktbarrett: These functions are being repeated in many places. Lift them into a common location and remove the duplicates.

pyuvm/_reg/uvm_mem.py
ysekhar: Fixed

pyuvm/_reg/uvm_mem.py
ysekhar: Fixed

pyuvm/_reg/uvm_reg_block.py
ysekhar: Fixed

pyuvm/uvm_reporting/__init__.py
```diff
@@ -25,6 +26,7 @@ def _parse_bool(value: Any) -> bool:
 
 
 _sv_uvm_style_reporting_enabled = _parse_bool(os.getenv(_ENV_VAR, "0"))
+_reg_model_logger = logging.getLogger("RegModel")
```
ktbarrett: This module isn't specific to the registers, so this doesn't seem correct.

I think something like `logging.getLogger(obj.name).warning(msg)` in the below functions would make more sense, but I'm not certain how well that would work.

ysekhar: There is something that does not work well with the suggestion in the overall reporting library. 
I am unable to debug/pinpoint it yet. 

For now I have moved all reg model reporting helpers into a shared module that caters for all pyral code. 
Will come back to debugging logging once all pyral work has upstreamed.

pyuvm/_reg/uvm_reg_reporting.py
```diff
@@ -0,0 +1,27 @@
...
+
+def uvm_reg_report_error(obj: Any, report_id: str, msg: str) -> None:
+    """Issue a RAL error through SV-UVM reporting or the RegModel logger."""
+    if get_sv_uvm_style_reporting_enabled():
+        obj.uvm_report.error(report_id, msg)
+    else:
+        _reg_model_logger.error(msg)
+
+
+__all__ = ["uvm_reg_report_error", "uvm_reg_report_warning"]
```
ktbarrett, line 27: Yeah this works. Thanks!

ktbarrett (approved):

PR #414, opened 2026-07-27
now with AI skills to review PRs

raysalemi:


ktbarrett:
Might be good to add 1800.2 and some of the pyuvm conventions where it breaks from the standard.

raysalemi:
Where would you put those?

ktbarrett:
In `references/` adjacent to `SKILL.md`. And you should put some notes in the SKILL pointing to it and when to use it.

PR #415, opened 2026-07-30
[pyral, core] Implement RAL sequence and frontdoor foundations

ysekhar:
# TL;DR

Implements one part of usable `uvm_reg_sequence` foundations, register and memory convenience helpers, layered register-transaction translation, and a constructible `uvm_reg_frontdoor` with cocotb task-owned locking.

This also removes the duplicate unusable `uvm_reg_frontdoor` placeholder and makes the sequence-derived implementation the authoritative top-level export.

# Summary

This PR builds on the register frontdoor foundation merged in #402 and the memory mapping and access support merged in #412.

It covers:

- constructible `uvm_reg_sequence` objects
- `model`, `adapter`, and `reg_seqr` sequence configuration
- both standard roles of `uvm_reg_sequence`:
  - a convenience base for user-defined register-model sequences
  - a layered register-to-bus translation sequence
- translation through the selected local map's `do_bus_write()` and `do_bus_read()` paths
- guaranteed upstream-item completion and parent restoration
- explicit validation of translation configuration and access kinds
- register convenience helpers:
  - `write_reg`
  - `read_reg`
  - `update_reg`
  - `mirror_reg`
- memory convenience helpers:
  - `write_mem`
  - `read_mem`
- target validation with centralized RAL diagnostics and standard failure results
- an awaited, Pythonic `uvm_reg_frontdoor.start()` implementation
- cocotb current-task ownership for frontdoor locking
- protection against non-owner unlocks and same-task lock reacquisition
- an explicit requirement for derived frontdoors to override `body()`
- top-level export of the sequence-derived `uvm_reg_frontdoor`
- explicit deferred-backdoor boundaries for register and memory peek/poke helpers

# RFC Alignment

This continues the staged RAL direction introduced in RFC #399 and builds on:

- #402: foundational register APIs and built-in register frontdoor access
- #412: memory mapping and built-in single-element memory frontdoor access

# Register Translation Sequences

`uvm_reg_sequence` now supports its layered translation role.

The default translation `body()`:

1. acquires a `uvm_reg_item` from the configured upstream register sequencer
2. temporarily installs the translation sequence as the item's parent
3. dispatches the operation through `do_reg_item()`
4. restores the original parent in `finally`
5. completes every acquired upstream item, including failure paths

`do_reg_item()` uses the item's selected local map and the sequence's configured downstream sequencer and adapter.

[...]

PR #418, opened 2026-08-01
[pyral, core] Support custom frontdoor 

ysekhar:
# TL;DR

Implements support of custom-frontdoor routing for register and memory operations.

Configured user-defined frontdoors now receive the complete originating `uvm_reg_item`, run on the correct sequencer, and return status and read data to the caller. The implementation supports mapped and intentionally unmapped elements, wider-than-bus custom accesses, shared-frontdoor serialization, and exception-safe state cleanup.

# Summary

This PR builds on the register and memory frontdoor foundations merged in #402, #412, and #415.

It covers:

- register read/write routing through configured custom frontdoors
- memory read/write routing through configured custom frontdoors
- complete frontdoor `rw_info` propagation
- explicit-frontdoor sequencer precedence
- operation-local root-map sequencer fallback
- useful diagnostics for missing or invalid sequencers
- status and read-data propagation from the frontdoor
- mapped custom-frontdoor access
- intentionally unmapped custom-frontdoor access
- rejection of unmapped elements without a custom frontdoor
- wider-than-bus memory access through user-defined frontdoors
- preservation of the built-in frontdoor’s single-beat width restriction
- register access-right validation
- memory access-right, offset, and mapping validation
- shared-frontdoor serialization
- exception-safe restoration of frontdoor sequencer state
- exception-safe release of frontdoor locks and register/memory busy state

# RFC Alignment

This continues the staged RAL direction introduced in RFC #399 and builds on:

- #402: foundational register APIs and built-in register frontdoor access
- #412: memory mapping and built-in single-element memory frontdoor access
- #415: register sequences, convenience helpers, translation sequences, and constructible frontdoors

# Custom Frontdoor Routing

When a register or memory operation resolves to a map entry with a configured frontdoor, the selected local map now dispatches the operation through that frontdoor rather than through the built-in adapter path.

The frontdoor receives the original `uvm_reg_item` through `rw_info`. This preserves:

- element and element kind
- access kind
- register or memory value
- logical memory offset
- selected map and local map
- access path
- parent sequence
- extension
- priority
- source filename and line number

The frontdoor updates the same item, so status and read data return naturally to the originating `read()` or `write()` call.

A normally returning frontdoor body counts as completion. Frontdoors control the final operation status through `rw_info`.

# Sequencer Selection

Custom-frontdoor sequencer selection follows this precedence:

1. use the sequencer explicitly configured on the frontdoor
2. otherwise use the selected map’s root-map sequencer for that operation
3. if neither is available, report a RAL configuration error and return `UVM_NOT_OK`

The root-map fallback is operation-local. It does not permanently replace the frontdoor’s configured sequencer.

The selected sequencer must be a usable `uvm_sequencer`. Invalid sequencer objects are diagnosed before the frontdoor body starts.

# Mapping and Width Behavior

Custom frontdoors support both mapped and intentionally unmapped register and memory entries.

An intentionally unmapped entry is accepted only when it has a configured custom frontdoor. Without one, the operation is rejected.

The built-in memory frontdoor retains its single-beat restriction when the memory element is wider than the map bus. A custom frontdoor may implement that wider physical access itself because its protocol and beat decomposition are owned by the user-defined sequence.

# Validation and Cleanup

Normal element validation remains active before custom-frontdoor execution.

This includes:

- register RO/WO map rights
- memory access rights
- memory offset bounds
- declared-width masking
- map association
- supported access path
- rejection of unmapped entries without a frontdoor

Shared frontdoors are serialized so concurrent operations cannot replace each other’s `rw_info` or sequencer state.

Cleanup is protected with `finally` blocks. If a user frontdoor raises or is cancelled:

- the operation is marked `UVM_NOT_OK`
- the original sequencer is restored
- the frontdoor lock is released
- register and memory in-progress state is cleared
- register busy state is cleared
- the original exception is re-raised

# Deliberate Scope Boundaries

This PR does not implement:

- register or memory backdoor dispatch
- `peek` or `poke`
- HDL-path access
- inherited block backdoors
- memory bursts
- generic built-in multi-beat transport
- predictor completion
- memory desired or mirrored state
- memory allocation management
- virtual registers or virtual fields
- RAL callbacks
- RAL coverage

# Tests Added

New focused coverage is provided by:

- `tests/pytests/reg/test_uvm_custom_frontdoor.py`

The tests cover:

- register custom-frontdoor reads and writes
- memory custom-frontdoor reads and writes
- built-in and custom-frontdoor result equivalence
- complete `rw_info` propagation
- parent sequence, extension, priority, and source propagation
- explicit-frontdoor sequencer precedence
- root-map sequencer fallback
- missing and invalid sequencers
- mapped custom frontdoors
- intentionally unmapped custom frontdoors
- rejection of unmapped entries without a frontdoor
- wider-than-bus custom memory access
- built-in width-limit preservation
- register RO/WO validation
- memory access-right and offset validation
- read-data masking
- frontdoor failure status
- exception propagation and cleanup
- shared-frontdoor concurrency
- memory frontdoor setter/getter behavior

### DIFF
--- a/pyuvm/_reg/uvm_mem.py
+++ b/pyuvm/_reg/uvm_mem.py
@@ -483,10 +483,10 @@ class uvm_mem(uvm_object):
             rw.set_status(uvm_status_e.UVM_NOT_OK)
             return None
         info = local_map.get_mem_map_info(self)
-        if info.unmapped or info.frontdoor is not None:
+        if info.unmapped and info.frontdoor is None:
             rw.set_status(uvm_status_e.UVM_NOT_OK)
             return None
-        if self.get_n_bytes() > local_map.get_n_bytes():
+        if info.frontdoor is None and self.get_n_bytes() > local_map.get_n_bytes():
             _report_error(
                 self,
                 "MEM_MULTIBEAT",
@@ -525,7 +525,10 @@ class uvm_mem(uvm_object):
         self._write_in_progress = True
         rw.set_status(uvm_status_e.UVM_IS_OK)
         try:
-            await rw.get_local_map().do_write(rw)
+            if info.frontdoor is None:
+                await rw.get_local_map().do_write(rw)
+            else:
+                await rw.get_local_map().do_frontdoor(rw, info.frontdoor)
         finally:
             self._write_in_progress = False
 
@@ -536,7 +539,10 @@ class uvm_mem(uvm_object):
         self._read_in_progress = True
         rw.set_status(uvm_status_e.UVM_IS_OK)
         try:
-            await rw.get_local_map().do_read(rw)
+            if info.frontdoor is None:
+                await rw.get_local_map().do_read(rw)
+            else:
+                await rw.get_local_map().do_frontdoor(rw, info.frontdoor)
             rw.set_value(rw.get_value() & ((1 << self._n_bits) - 1))
         finally:
             self._read_in_progress = False
@@ -548,10 +554,23 @@ class uvm_mem(uvm_object):
         fname: str = "",
         lineno: int = 0,
     ) -> None:
-        raise NotImplementedError
+        self._fname = fname
+        self._lineno = lineno
+        local_map = self.get_local_map(map)
+        if local_map is None:
+            return
+        info = local_map.get_mem_map_info(self)
+        if info is not None:
+            info.frontdoor = ftdr
 
     def get_frontdoor(self, map: uvm_reg_map = None) -> uvm_reg_frontdoor:
-        raise NotImplementedError
+        local_map = self.get_local_map(map)
+        if local_map is None:
+            return None
+        info = local_map.get_mem_map_info(self)
+        if info is None:
+            return None
+        return info.frontdoor
 
     def set_backdoor(
         self, bkdr: uvm_reg_backdoor, fname: str = "", lineno: int = 0
--- a/pyuvm/_reg/uvm_mem.py
+++ b/pyuvm/_reg/uvm_mem.py
@@ -563,7 +563,7 @@ class uvm_mem(uvm_object):
         if info is not None:
             info.frontdoor = ftdr
 
-    def get_frontdoor(self, map: uvm_reg_map = None) -> uvm_reg_frontdoor:
+    def get_frontdoor(self, map: uvm_reg_map = None) -> uvm_reg_frontdoor | None:
         local_map = self.get_local_map(map)
         if local_map is None:
             return None
--- a/pyuvm/_reg/uvm_reg_map.py
+++ b/pyuvm/_reg/uvm_reg_map.py
@@ -1011,7 +1011,7 @@ class uvm_reg_map(uvm_object):
 
     async def do_frontdoor(
         self, rw: uvm_reg_item, frontdoor: uvm_reg_frontdoor
-    ) -> bool:
+    ) -> None:
         prior_sequencer = frontdoor.sequencer
         sequencer = prior_sequencer
         if sequencer is None:
@@ -1023,22 +1023,14 @@ class uvm_reg_map(uvm_object):
                 if element is not None and hasattr(element, "get_full_name")
                 else rw.get_name()
             )
-            _report_error(
-                self,
-                "REG_FRONTDOOR",
+            raise UVMFatalError(
                 f"Custom frontdoor access to {element_name!r} has no sequencer; "
-                "configure the frontdoor sequencer or the root-map sequencer",
+                "configure the frontdoor sequencer or the root-map sequencer"
             )
-            rw.set_status(uvm_status_e.UVM_NOT_OK)
-            return False
         if not isinstance(sequencer, uvm_sequencer):
-            _report_error(
-                self,
-                "REG_FRONTDOOR",
-                f"Custom frontdoor sequencer {sequencer!r} is not a uvm_sequencer",
+            raise UVMFatalError(
+                f"Custom frontdoor sequencer {sequencer!r} is not a uvm_sequencer"
             )
-            rw.set_status(uvm_status_e.UVM_NOT_OK)
-            return False
 
         await frontdoor.atomic_lock()
         try:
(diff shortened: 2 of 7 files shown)


raysalemi:
Some of the regression tests are failing.

Other than that, Gemini is very happy:
```
Viewed SKILL.md:1-51
Viewed README.md:1-713
Ran command: `gh pr view 418 --json title,body`
Ran command: `gh pr diff 418`
Ran command: `gh pr view 418 --json title,body`
Ran command: `gh pr diff 418`
Ran command: `gh pr diff 418 | wc -l`
Ran command: `gh pr diff 418 > /Users/raysalemi/repos/github_pyuvm/pr418_diff.txt`
Viewed pr418_diff.txt:1-800
Viewed pr418_diff.txt:800-856
```
Here is a review of **Pull Request 418** based on the PR Reviewer Skill criteria:

### Overview
This pull request implements support for custom frontdoors for register and memory operations, allowing users to define their own access sequences. The PR is approximately 850 lines of diff and introduces significant changes to routing, validations, error handling, and concurrency management for `uvm_reg` and `uvm_mem`. 

### Test Sufficiency Matrix
The code changes introduce many new conditional branches and exception handling scenarios, all of which are exhaustively tested.
* **Conditional Routing:** Tests (`test_register_custom_frontdoor_write_and_read_match_bus_results`, `test_memory_custom_frontdoor_write_and_read_match_bus_results`) verify that both registers and memories correctly route to a custom frontdoor when present.
* **Sequencer Selection & Diagnostics:** The fallback logic from an explicitly configured frontdoor sequencer to the root-map sequencer is tested (`test_explicit_frontdoor_sequencer_precedes_root_map_sequencer`). Edge cases for missing or invalid sequencers are successfully caught and tested (`test_missing_custom_frontdoor_sequencer_reports_failure`, `test_invalid_custom_frontdoor_sequencer_is_rejected_before_activity`).
* **Unmapped & Wide Elements:** Logic allowing intentionally unmapped elements (when a frontdoor is provided) and rejecting them (when missing) is tested. Tests also confirm that a custom frontdoor can bypass the single-beat width limit placed on built-in access.
* **Validations:** Access rights (`RO` vs `WO`) and out-of-bounds checks correctly execute *before* custom frontdoor activity.
* **Concurrency & Exceptions:** The newly added `try/finally` blocks and `atomic_lock()` locking in `do_frontdoor` are validated by `test_frontdoor_exception_restores_state_and_is_reraised` and `test_shared_frontdoor_serializes_concurrent_memory_operations`. These correctly confirm that locks and busy-states are released upon exceptions.

**Coverage Conclusion:** Excellent. Every new branch, exception path, and validation step introduced in the diff has corresponding test coverage in the new `test_uvm_custom_frontdoor.py` suite.

### Intent Realization Alignment
The changes cleanly align with the stated goals in the PR description:
* **Custom-Frontdoor Routing:** Read and write logic successfully routes through `local_map.do_frontdoor` rather than the built-in adapter path. 
* **Complete `rw_info` Propagation:** The frontdoor properly receives the original `uvm_reg_item`, retaining offsets, data, and access paths, and returns statuses back to the caller.
* **Exception-Safe Cleanup:** `finally` blocks in `uvm_reg.py`, `uvm_mem.py`, and `uvm_reg_map.py` guarantee that `_write_in_progress`, `_read_in_progress`, busy states, locks, and temporary sequencer overriding states are safely cleaned up. 
* **Scope Control:** The PR explicitly avoids generic built-in multi-beat transport, backdoors, and bursts, successfully staying within the deliberate boundaries mentioned in the PR description.

### Recommendation
**PASS** 

The PR perfectly realizes its stated intent, provides thorough test coverage of both the happy paths and the edge cases (exceptions, invalid config, concurrency), and cleanly integrates into the existing UVM RAL structure.

pyuvm/_reg/uvm_reg_map.py
```diff
@@ -1007,6 +1009,50 @@ async def do_read(self, rw: uvm_reg_item) -> None:
...
+            rw.set_status(uvm_status_e.UVM_NOT_OK)
+            return False
+        if not isinstance(sequencer, uvm_sequencer):
+            _report_error(
+                self,
+                "REG_FRONTDOOR",
+                f"Custom frontdoor sequencer {sequencer!r} is not a uvm_sequencer",
+            )
+            rw.set_status(uvm_status_e.UVM_NOT_OK)
+            return False
```
ktbarrett: reporting an error and then returning a error boolean sounds like it should be an exception instead.

pyuvm/_reg/uvm_mem.py
```diff
@@ -548,10 +554,23 @@ def set_frontdoor(
...
+        self._fname = fname
+        self._lineno = lineno
+        local_map = self.get_local_map(map)
+        if local_map is None:
+            return
+        info = local_map.get_mem_map_info(self)
+        if info is not None:
+            info.frontdoor = ftdr
 
     def get_frontdoor(self, map: uvm_reg_map = None) -> uvm_reg_frontdoor:
```
ktbarrett:
```suggestion
    def get_frontdoor(self, map: uvm_reg_map = None) -> uvm_reg_frontdoor | None:
```

pyuvm/_reg/uvm_reg_map.py
ysekhar: This give an option to the sequence to restructure as needed. Since uvm_reporting has message and severity demotion this is handled differently. The error can be downgraded to an info in very specific cases.
 
May be this is an exception. Will require some downstream code to change from what people are used when migrating from SV-UVM. 

For now I have fixed this with no return but the status can be used for decisions.

pyuvm/_reg/uvm_mem.py
ysekhar: Not sure why this changed. Thanks for the spot.

pyuvm/_reg/uvm_reg_map.py
raysalemi: We should default to Pythonic behavior. We should not expect the people using UVM reporting beyond what is already in pyuvm's basic reporting .

raysalemi: I agree with Kaleb on throwing exceptions in this case

ysekhar: Done. Changed this to an Exception.

ktbarrett (approved):

PR #422, opened 2026-08-06
feat: add non-blocking sequence item polling

### LINKED ISSUE #421: Add try_next_item
### Description

IEEE 1800.2 UVM defines `try_next_item()` as a non-blocking method on `uvm_seq_item_pull_port` to check for available sequence items without waiting. We need to implement and expose this method on `seq_item_port`.

### Tasks

* Implement `try_next_item()` on `seq_item_port`.
* Add unit tests in `tests/` covering both available and empty sequence queue cases.
* Verify existing driver/sequencer tests continue to pass.

### Contribution Guidelines

Please include full unit test coverage for any new functionality added to `pyuvm`. Pull requests without corresponding tests cannot be merged.

aryansk:
## Summary
- Add non-blocking `try_next_item()` support to the sequence-item port/export.
- Return a (success, item) tuple while preserving current-item and completion semantics.
- Add connected-port tests for available and empty queues.

Fixes https://github.com/pyuvm/pyuvm/issues/421

## Validation
- `uv run pytest` (605 passed, 7 expected failures)
- `uv run ruff check`

### DIFF
--- a/pyuvm/_s12_uvm_tlm_interfaces.py
+++ b/pyuvm/_s12_uvm_tlm_interfaces.py
@@ -100,6 +100,7 @@ class uvm_port_base(uvm_export_base):
         "put_req",
         "put_response",
         "get_next_item",
+        "try_next_item",
         "item_done",
         "get_response",
     ]
--- a/pyuvm/_s14_15_python_sequences.py
+++ b/pyuvm/_s14_15_python_sequences.py
@@ -7,6 +7,7 @@
 # SystemVerilog features.
 
 
+from cocotb.queue import QueueEmpty
 from cocotb.triggers import Event as CocotbEvent
 
 from pyuvm._error_classes import UVMFatalError, UVMSequenceError
@@ -227,6 +228,24 @@ class uvm_seq_item_export(uvm_blocking_put_export):
         await self.current_item.item_ready.wait()
         return self.current_item
 
+    def try_next_item(self):
+        """Return the next queued sequence item without waiting.
+
+        ``None`` means the request queue is empty. Like ``get_next_item``, this
+        method tracks the returned item until ``item_done`` is called.
+        """
+        if self.current_item is not None:
+            raise UVMSequenceError(
+                "You must call item_done() before calling try_next_item again"
+            )
+        try:
+            self.current_item = self.req_q.get_nowait()
+        except QueueEmpty:
+            return None
+        self.current_item.start_condition.set()
+        self.current_item.start_condition.clear()
+        return self.current_item
+
     def item_done(self, rsp=None):
         """
         Signal that the item has been completed. If ``rsp`` is not ``None``
@@ -300,6 +319,14 @@ class uvm_seq_item_port(uvm_port_base):
         """
         try:
             return await self.export.get_next_item()
+        except AttributeError:
+            assert self.export is not None, "export is not connected"
+            raise
+
+    def try_next_item(self):
+        """Return the next sequence item immediately, or ``None`` if empty."""
+        try:
+            return self.export.try_next_item()
         except AttributeError:
             assert self.export is not None, "export is not connected"
             raise
--- a/pyuvm/_s14_15_python_sequences.py
+++ b/pyuvm/_s14_15_python_sequences.py
@@ -229,10 +229,11 @@ class uvm_seq_item_export(uvm_blocking_put_export):
         return self.current_item
 
     def try_next_item(self):
-        """Return the next queued sequence item without waiting.
+        """Return ``(success, item)`` for the next queued sequence item.
 
-        ``None`` means the request queue is empty. Like ``get_next_item``, this
-        method tracks the returned item until ``item_done`` is called.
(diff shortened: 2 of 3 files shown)


raysalemi:
Thanks for picking this up so quickly!

We should make this try work the same way as `try_get()` The `try_get()` function returns a tuple with a success boolean:

```
    def try_get(self):
        """
        :raises: UVMTLMConnectionError if export is missing
        :return: (success, data)

        12.2.4.2.6
        Returns a tuple containing success and the data
        This is different than SV UVM that returns the
        data through the argument list.
        :return: (success, data)
        """

        try:
            (success, data) = self.export.try_get()
        except AttributeError:
            raise UVMTLMConnectionError(
                "Missing or wrong export in"
                f" {self.get_full_name()}. Did you connect it?"
            ) from None

        return success, data

```

raysalemi:
Unless the standard says this is different.  In that case, we should point to the standard.

aryansk:
Thanks for the clarification. I updated try_next_item to follow try_get and try_peek: it now returns (True, item) when an item is available and (False, None) when the queue is empty. I also updated the connected-port regression tests. Commit 50b89a6 is pushed, and the targeted pytest and ruff checks pass. I left the PR in draft for further review.

aryansk:
Follow-up validation: the full uv run pytest suite now passes with 605 passed and 7 expected xfails.

raysalemi:
Thank you!

It just needs two more tests:

### 1. Test for `UVMSequenceError` on repeated calls without `item_done()`

In `uvm_seq_item_export.try_next_item()`, there is a check to ensure `try_next_item()` isn't called again until the current item finishes:
```python
if self.current_item is not None:
    raise UVMSequenceError(
        "You must call item_done() before calling try_next_item again"
    )
```
We need a test to ensure this safeguard actually triggers if someone misuses the API. You can drop this into `tests/pytests/test_uvm_seq_item_port.py`:

```python
from pyuvm._error_classes import UVMSequenceError

def test_try_next_item_raises_error_if_called_twice_without_item_done():
    port, export = make_connected_port()
    item = uvm_sequence_item("item")
    export.req_q.put_nowait(item)

    # First call succeeds
    success, result = port.try_next_item()
    assert success is True

    # Second call without item_done() should raise UVMSequenceError
    with pytest.raises(UVMSequenceError, match="You must call item_done\\(\\) before calling try_next_item again"):
        port.try_next_item()
```

### 2. Test for Disconnected Port 

In `uvm_seq_item_port.try_next_item()`, there's an exception handler that catches `AttributeError` and raises an assertion if the port is disconnected:
```python
except AttributeError:
    assert self.export is not None, "export is not connected"
    raise
```
We need a test to verify that calling `try_next_item()` on a disconnected port properly raises this assertion.

```python
def test_try_next_item_raises_assertion_error_if_export_is_disconnected():
    root = uvm_root()
    # Create a port but DO NOT connect it to an export
    port = uvm_seq_item_port("seq_item_port", root)

    with pytest.raises(AssertionError, match="export is not connected"):
        port.try_next_item()
```

Thanks again!

aryansk:
Added both requested regression tests in commit 856f6bf:

- repeated try_next_item() without item_done() raises UVMSequenceError with the documented message;
- a disconnected port raises AssertionError with export is not connected.

Validation: the targeted file passes Ruff, the focused test file passes 4/4, the full suite passes 607 tests with 7 expected xfails, and git diff --check passes.

raysalemi:
# PR 422 Review (Updated)

## PR Details
**Title:** feat: add non-blocking sequence item polling
**Description Summary:** Adds non-blocking `try_next_item()` support to the sequence-item port/export, returning a `(success, item)` tuple.

## Size Check
**Pass:** The PR diff is 123 lines, which is well under the 1000-line limit.

## Intent Realization Alignment
**Pass:** The PR successfully implements the goals stated in the title and body.
* It adds `try_next_item()` to the sequence-item port and export.
* It returns a `(success, item)` tuple.
* It preserves the current-item and completion semantics.
* It adds connected-port tests for the available and empty queues.

## Test Sufficiency Matrix
**Pass:** The code adds logic paths that are all correctly covered by tests. 
The contributor successfully added the tests for the two edge cases:
1. **`UVMSequenceError` path:** `test_try_next_item_raises_error_if_called_twice_without_item_done` verifies that repeated calls without `item_done()` raise the correct exception.
2. **Disconnected Port path:** `test_try_next_item_raises_assertion_error_if_export_is_disconnected` verifies that calling `try_next_item()` on a disconnected port raises the correct `AssertionError`.

## Conclusion and Recommendation
**Recommendation: Approve (Pass)**

The PR intent is fully realized and the test coverage is now robust, handling both the happy paths and the error edge cases properly. It is ready to be merged!

raysalemi:
Great job.  Thank you!

aryansk:
Thanks for the updated review and the kind words, @raysalemi. I appreciate you checking both edge cases against the implementation itself. It is great to have the repeated-call and disconnected-port paths confirmed before the PR landed.

